PR #37, opened 2021-12-02
improvement to the check_dns

ftmiranda:
This code, will verify the dns configuration, not only test if the FQDNs resolves to IP. It will also verify that each FQDN for each node really matches the node's IP defined into the env.yaml
This is not 100% my contribution, Jacob heavily contributed to this code as well.


jacobemery (approved):
Looks good!

PR #38, opened 2021-12-13
ssh-copy-id: make expect script idempotent

NicoBoehrIBM:
This makes the expect script idempotent, if the SSH key was already
present on the machine.


jacobemery (changes requested):
Works great, just tested it out. Just needs to be updated to use new variables, since original submission:

ssh_target[2] instead of env.access.login.kvm.root_password

Thanks Nico!

jacobemery:
Nico gave the go-ahead to merge. My own tests proved it works perfectly. Thanks Nico!

PR #39, opened 2022-03-27
Automated KVM Provision & Install Update

jacobemery:
Now supports using the zHMC Ansible modules to create an LPAR profile, attach storage group and network cards, boot RHEL from FTP server, and kickstart the installation. Various other changes and improvements. The biggest changes come from aligning with Ansible best practices. See CHANGELOG for more details.


jacobemery:
Will continue to test and improve upon this update in the coming weeks. Next step is to create a VPN so that the number of IPs needed decreases, and so that 3 LPARs can be used (for high availability) and can communicate with each other through the private network.

PR #42, opened 2022-12-21
Fix for the default storage pool usage. Use newly created storage pool.

AmadeusPodvratnik:
During installation an own storage pool is being created named: {{ env.cluster.networking.metadata_name }}-vdisk
During installation of following nodes: bootstrap, control, compute and infra the default pool is being used rather than the newly created storage pool. 
This may lead to errors during execution of the playbook scripts because the file path of default storage pool may differ from storage path for newly created pool. 
This fix uses the newly created ({{ env.cluster.networking.metadata_name }}-vdisk) pool instead of the default pool.


jacobemery (approved):
Thanks for catching this! Looks good to me.

PR #43, opened 2022-12-21
Enable root user as installing user

AmadeusPodvratnik:
Sometimes it is necessary (and very convenient) to work with the root user instead of using a specific user with sudo access.
Especially when used in a closed test environment in which various automation tests are being executed. 
This fix enables root user support for edge use cases.


jacobemery:
Added comment in docs to help users when setting password variables. And updated some of the descriptions. Otherwise looks good to me, thanks Amadeus!!

jacobemery (approved):
LGTM

PR #44, opened 2023-01-04
Fix typo in tag

silliman:
Changed 'virt-isntall' to 'virt-install' on line 61  (Find FTP home for use in next step)

Signed-off-by: Barry Silliman silliman@us.ibm.com


jacobemery (approved):
Good catch, thanks Barry!

PR #48, opened 2023-01-09
feat: Add .editorconfig

smolin-de:
Add .editorconfig to have consistent coding styles across editors and IDEs.


jacobemery (approved):
Great idea!

PR #51, opened 2023-01-09
feat: Untrack all.yaml file in git

smolin-de:
The inventory file 'all.yaml' is part of the current git structure. This file is always modifed to execute Ansible-OpenShift-Provisioning and therefore it will be tracked in your git branch as modifed. But, it will not be commited, because it contains credentials !

This patch renames the original file to a template filename. The user needs to create a local 'all.yaml' copy.

### DIFF
--- a/docs/set-variables-group-vars.md
+++ b/docs/set-variables-group-vars.md
@@ -1,7 +1,7 @@
 # Step 2: Set Variables (group_vars)
 ## Overview
-* In a text editor of your choice, open the [environment variables file](https://github.com/IBM/Ansible-OpenShift-Provisioning/blob/main/inventories/default/group_vars/all.yaml).
-* This is the master variables file and you will likely reference it many times throughout the process. The default inventory can be found at [inventories/default/group_vars/all.yaml](https://github.com/IBM/Ansible-OpenShift-Provisioning/blob/main/inventories/default/group_vars/all.yaml).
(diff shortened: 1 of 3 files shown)


smolin-de:
> Definitely agree on adding all.yaml to untrack git files.
> 
> What is your reasoning for the template for all.yaml? Just to have a clean copy if they need to start fresh? Maybe instead of requiring them to create a local copy, we just create a clean backup for them instead. That way it's one less step right at the beginning.
> 
> What do you think @smolin-de ?

> What is your reasoning for the template for all.yaml?
all.yaml is part of the git and it must be modified. I have always uncommitted git changes. When I work with other git branches, then I need to stage my changes locally or delete the all.yaml file. I need be very careful to not accidentally push my changes to the git tree where everybody can see them, because all.yaml contains some of my private credentials
I just want that my config file (with credentials) is not part of the got tree.

jacobemery (commented):
> I need be very careful to not accidentally push my changes to the git tree

Ah I see. I figured there'd be a way to ignore future changes to a file with GitHub, but I'm not finding one. This sounds like a fine workaround.

It does add a step at the beginning, but that's better than accidentally exposing credentials. Plus, this will prime users to think about the names of the inventory files for when they will have to change the name(s) of the KVM host inventory files as well.

I'll have to add a short step to the docs as well. Let me get that ready so I can push that and merge this at relatively the same time.

jacobemery:
Added documentation, good to merge.

jacobemery (approved):
Definitely need this.

PR #52, opened 2023-01-10
fix: Problem wit ansible_python_interpreter variable and value

smolin-de:
The default ansible variable for python is 'interpreter_python=' not 'ansible_python_interpreter=' Default path on a MacOS and RHEL is '/usr/bin/python3'. Use '/usr/bin/env python3' to resolve the path issue on other OS.


jacobemery:
Great idea, let me do a quick test. Otherwise should be good to go!

jacobemery (commented):
When I run this the reference to {{ interpreter_python }} in roles/set_inventory/templates/hosts.j2 comes back with an undefined error since the only place it's defined is in ansible.cfg. 

But with this new method for setting the default interpreter_python value in ansible.cfg to a more universal value, may make setting the python interpreter for the workstation in the template obsolete anyway. I'll test this out.

jacobemery (approved):

PR #53, opened 2023-01-10
feat: Use KVM host to store bastion kickstart file

smolin-de:
The bastion kickstart file is stored on a ftp server and then dowloaded from the KVM host. A ftp root password is required to verify http service. We do not have always the root password for an existing ftp server. This patch simplifies the setup and stores the bastion kickstart file on the KVM host.


jacobemery (approved):
Tested, much faster and simpler. Nice work!

jacobemery:
Great work!! I just had to add rhel to os-variant to avoid an error with virt-install.

Using the KVM host is so much faster and reduces complexity with http.

PR #55, opened 2023-01-10
fix: Remove unused spaces

smolin-de:


jacobemery:
In testing this out, I found an error in what I pushed for the proxy info in the install-config template.

jacobemery (approved):
Thanks for doing this!

PR #56, opened 2023-01-11
fix(ssh): Some minor fixes

smolin-de:
fix: Set timeout for ssh-copy-id, do not wait forever
fix: Force ssh-copy-id, to avoid issues if the key already exists
fix: Duplicate creation of 'path_to_key_pair' variable in 'all.yaml' file


jacobemery (approved):
Looks good to me!

PR #57, opened 2023-01-11
fix: Add retry if subscription server is not available

smolin-de:


jacobemery (approved):
Nice fail safe, subscription manager sometimes has downtime.

PR #59, opened 2023-01-12
fix(bootstrap): Deletion of existing VM image

smolin-de:
Delete correct bootstrap qcow2 image.

The bootstrop VM is deleted after creation of control nodes. But, in case of an issue, the VM is not deleted. New config bootstrap changes are ignored, because the existing old bootstrap VM is used on the next executions until the VM is manually deleted.
We are now deleting and creating the bootstrap VM on each test run, to solve the above issue.


jacobemery (approved):
Tested, works great!

PR #60, opened 2023-01-12
feat: Add delete_nodes role

smolin-de:
In case you want to re-install the cluster, because of an install problem or of a new OCP version, then you can just execute playbooks:
 * 6_create_nodes.yaml
 * 7_ocp_verification.yaml

In order to have a clean setup on each run, we delete all existing control, compute and infra nodes. Bastion node will not be deleted.


jacobemery (approved):
Tested, works great!

PR #61, opened 2023-01-12
fix: Improve output of openshift-installer progress and check

smolin-de:
The openshift-installer requires up to 60 min to complete. You need to check manually if the install process started sucessfully or you simply wait 60 min for the results.
This patch monitors the 'node-bootstrapper' request. Usually such request is raised within 6 min. If not, then this is considered as an error and the ansible playbook script exits.

The openshift-installer process is not aborted! It is still running.

### DIFF
--- a/roles/wait_for_bootstrap/tasks/main.yaml
+++ b/roles/wait_for_bootstrap/tasks/main.yaml
@@ -1,30 +1,58 @@
 ---
 
-- name: Watch wait-for bootstrap-complete process.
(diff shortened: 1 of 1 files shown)


jacobemery (approved):
Tested it out and it looks great! I just had to add the lowercase filter just in case the user sets their hostnames in all.yaml in all caps (like I did), which caused it to timeout even though the value the task was checking for was present.

PR #62, opened 2023-01-13
fix: Enable autostart for control, compute and infra VMs

smolin-de:


jacobemery:
This is great @smolin-de! Could you add a commit to this PR for the bootstrap too? It doesn't have autostart on either and we might as well get them all done at the same time.

jacobemery (approved):
Great idea! I added a commit for the bootstrap as well.

PR #63, opened 2023-01-16
Revert bastion nameserver patch

smolin-de:
The required parameter "--extra-args" was removed in the virt-install command.
This is causing the VM creation to fail. 
Therefore removing the patch until a new patch is available.


jacobemery (approved):
Sorry about that! I tested and it worked, not sure how that slipped by me. Thanks for catching it.

PR #65, opened 2023-01-16
Add NAT support

AmadeusPodvratnik:
This code change add NAT support to the code. In case of KVM NAT network mode and an installation from remote (e.g. from your laptop). In case of NAT the IP of the bastion and cluster nodes will use 192.168.x.x adr. which are usually not accessible from remote. To access those addresses a ssh tunnel is needed where usually the KVM host acts as a jumphost. 
This new features is adding new variables to the all.yaml to specify the network mode (NAT) and the required variables for hostname, user and password. If network mode set to NAT a ssh config file is being created containing the a bastion node and the matching jumphost.
In addition a small condition fix will added, too (issue with root user and the correct home directory).


jacobemery:
Testing this tomorrow. Much needed! Thanks for this @AmadeusPodvratnik

jacobemery (changes requested):
I appreciate that you added info to the docs! Could you add a relevant section to the all.yaml.template file and commit it?

jacobemery:
Does this require a change to the macvtap role @AmadeusPodvratnik ? For example, when network mode is NAT, don't run macvtap role when setting up the KVM host?

jacobemery:
When using NAT, would we want to edit the xml of the default network to exclude automatic allocation of a range of IPs corresponding to the number of nodes? So that the IPs used for the nodes are not automatically allocated to a new VM, creating an IP conflict. @AmadeusPodvratnik

AmadeusPodvratnik:
Hi @jacobemery, thanks for the comments and suggestions. 
Good catch for macvtap network interface. I modified 3_setup_kvm_host.yaml and enhance the documentation. A possible conflict must be resolved by the user (virsh net-edit) to prevent conflicts with dhcp and the OCP nodes.
Add the new vars to the template file (not sure why I missed that).

PR #66, opened 2023-01-16
fix: Problem with hostname and 2nd nameserver definition

smolin-de:
Hostname was not correctly defined in ks
nameserver2 variable was not correctly used in ks


jacobemery (approved):
Good catch, thanks!

PR #67, opened 2023-01-16
synchronous the template files

routerhan:
fix: only the `all.yaml` shall be ignored because it contains credentials. 
fix: update the template file with the latest renaming: `workstation` to `controller`

Signed-off-by: Ching Han Chen <ching.chen@ibm.com>

### DIFF
--- a/inventories/default/group_vars/.gitignore
+++ b/inventories/default/group_vars/.gitignore
@@ -1,3 +1,2 @@
 /*
-!.gitignore
-!all.yaml.template
+!all.yaml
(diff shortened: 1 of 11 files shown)


routerhan:
From my point of view, the template file should not be included in the gitignore, as we are copying this file as template for further configuration. Keeping it up to date is required.

jacobemery:
@routerhan The exclamation point ! in the .gitignore file means 'except', so ignore everything in the folder (/*) _except_ the template (all.yaml.template) and the .gitignore file itself. This suggested change would remove those two exceptions (making them untracked) and add an exception for all.yaml (making it tracked). That's the opposite of what we want (ignore all.yaml and track the other two).

The other commit (changing the template to controller instead of workstation) is needed though, could you separate these commits so we can pull just that one?

routerhan:
I will close this one and open another PR... as the commits is messy.

PR #68, opened 2023-01-17
fix: Improve waiting for cluster operators check

smolin-de:
Replace seven static tasks to check cluster operators, with a task loop.


jacobemery (approved):
Tested, works beautifully! Nice work. I like how it shows only the cluster operators that are in 'progressing' state and then does all of them at the end. Really well done!

PR #70, opened 2023-01-18
fix: Add api.<metadata_name>.<base_domain> to /etc/hosts

smolin-de:
Add api.{{ metadata_name }}.{{ base_domain }} to /etc/hosts Remove 'wait for install check'. Because install is already completed at this stage


jacobemery (approved):
lgtm, thanks!

PR #74, opened 2023-01-18
fix: Improve node check

smolin-de:
Verify the status of all defined control and compute nodes. The old code just checked if reported nodes where 'Ready'. The reported node list was not verified.


jacobemery (approved):
I was going to open an issue for this exact thing, thanks for doing this!

Just had to add a lowercase filter for users that enter their hostnames in capital letters.

PR #75, opened 2023-01-18
Add reinstall_cluster playbook

jacobemery:
Since the get_ocp role refreshes ignitions already, the clean_ocp_rhcos_inst_files role is not needed.

And since the reinstall_nodes playbook that Klaus added didn't refresh the ignitions, as would be needed if the cluster was over 24 hours old, I combined these two ideas into a new playbook reinstall_cluster.yaml.

I also updated the docs. I think having the 5.1 playbook in the middle of the required playbooks for installation is a bit confusing, so I created a new section for additional playbooks and added this and the other playbooks there.

I hope this all makes sense and is ok by you @AmadeusPodvratnik! This capability was needed, and this was a great way to re-visit it.


jacobemery (commented):
@AmadeusPodvratnik Please confirm these changes

AmadeusPodvratnik (approved):
Review done with @jacobemery.

PR #76, opened 2023-01-19
fix: Create nodes on LPAR1 in non high availability mode

smolin-de:


jacobemery (approved):
lgtm, just added one for infra as well.

PR #77, opened 2023-01-20
feat: Add NAT support beside macvtap

AmadeusPodvratnik:
This code change add NAT support to the code. In case of KVM NAT network mode and an installation from remote (e.g. from your laptop). In case of NAT the IP of the bastion and cluster nodes will use 192.168.x.x adr. which are usually not accessible from remote. To access those addresses a ssh tunnel is needed where usually the KVM host acts as a jumphost.
This new features is adding new variables to the all.yaml to specify the network mode (NAT) and the required variables for hostname, user and password. If network mode set to NAT a ssh config file is being created containing the a bastion node and the matching jumphost.
In addition a small condition fix will added, too (issue with root user and the correct home directory).


AmadeusPodvratnik:
Hi Jacob, I just created a new pull request have the items we discussed yesterday (hopefully I did not missed something). I will delete the old pull request.

roles/ssh_add_config/tasks/main.yaml
```diff
@@ -0,0 +1,47 @@
...
+    path: ~/.ssh/config
+    backup: true
+    create: true
+    mode: '0644'
+    block: |
+      Host bastion-jump
+        HostName {{ env.jumphost.name }}
+        User {{ env.jumphost.user }}
+        IdentityFile ~/.ssh/id_rsa
+      #Host bastion-{{ env.jumphost.name }}
```
smolin-de, line 19: delete comment !

smolin-de, line 19:
```suggestion
```

roles/set_inventory/templates/hosts.j2
```diff
@@ -11,3 +11,6 @@
 
 [bastion]
 {{ env.bastion.networking.hostname }} ansible_host={{ env.bastion.networking.ip }} ansible_user={{ env.bastion.access.user }} ansible_become_password={{ env.bastion.access.pass }}
+
+[jumphost]
+{{ env.jumphost.name }} ansible_host={{ env.jumphost.ip }} ansible_user={{ env.jumphost.user }} ansible_become_password=i{{ env.jumphost.pass }}
```
jacobemery, line 16:
```suggestion
{%- if env.jumphost.name is not none %}
{{ '[jumphost]' }}
{{ env.jumphost.name | string + ' ansible_host=' + env.jumphost.ip | string + ' ansible_user=' + env.jumphost.user | string + ' ansible_become_password=' + env.jumphost.pass | string }}
{%- endif %}
```

jacobemery: This jinja conditional formatting helps avoid some potential errors. This is how it works:
- The [jumphost] section will only be added if env.network_mode is NAT and all the jumphost vars are defined. 
- Ensures all jumphost vars are concatenated to strings to avoid an error if someone sets env.jumphost.name to an integer, for example. 
- It also makes sure evaluation of env.network_mode doesn't fail if the value is lowercase.

docs/set-variables-group-vars.md
```diff
@@ -167,4 +167,10 @@
...
+**env.network_mode: | (Optional) In case the network mode will be NAT and the installation will be executed from remote (e.g. your laptop), a jumphost needs to be defined to let the installation access the bastion host. If macvtap for networking is being used this variable should be empty.
+**env.jumphost.name: | (Optional) If env.network.mode is set to 'NAT' the name of the jumphost (e.g. the name of KVM host if used as jumphost) should be specified.
+**env.jumphost.ip: | (Optional) The ip of the jumphost.
+**env.jumphost.user: | (Optional) The user name to login to the jumphost.
+**env.jumphost.pass: | (Optional) The password for user to login to the jumphost.
+**env.jumphost.path_to_keypair: | The path to the keypair on the jumphost to be copied to the bastion.
```
jacobemery, line 176:
```suggestion
**env.jumphost.path_to_keypair: | (Optional) The path to the keypair on the jumphost to be copied to the bastion.
```

roles/set_inventory/templates/hosts.j2
jacobemery: Also removes 'i' typo after 'ansible_become_password='

playbooks/5_setup_bastion.yaml
```diff
@@ -1,28 +1,29 @@
...
+  become: true
   gather_facts: false
   vars_files:
     - "{{ inventory_dir }}/group_vars/all.yaml"
   vars:
-    ssh_target: ["{{ env.bastion.networking.ip }}", "{{ env.bastion.access.user }}", "{{ env.bastion.access.pass }}"]
+    ssh_target: ["{{ env.bastion.networking.ip }}", "{{ env.bastion.access.user }}", "{{ env.bastion.access.pass }}", "{{ env.jumphost.path_to_keypair }}", "{{ env.jumphost.ip }}"]
   roles:
-    - ssh_copy_id
+    - { role: ssh_add_config, tags: ssh_copy_id, ssh, when: env.network_mode == "NAT" }
```
jacobemery, line 13:
```suggestion
    - { role: ssh_add_config, tags: ssh_copy_id, ssh, when: env.network_mode | upper == "NAT" }
```

jacobemery: ensure that the task won't fail if users set env.network_mode to lowercase 'nat' instead of 'NAT'

jacobemery (changes requested):
Add a failed_when conditional to [roles/set_inventory/tasks/main.yaml](https://github.com/IBM/Ansible-OpenShift-Provisioning/blob/main/roles/set_inventory/tasks/main.yaml), after line 21, which fails the task when jumphost vars are undefined and env.network_mode | upper == 'NAT' to avoid the case where users forget to define all the necessary jumphost vars. That way they know immediately, instead of silently skipping past it or entering it incompletely and failing later and not knowing the cause.

```
- name: Template out inventory with localhost, FTP, KVM host, and bastion information
  tags: set_inventory
  template:
    src: hosts.j2
    dest: "{{ find_project.stdout }}{{ find_inventory.stdout }}/hosts"
    force: yes
  failed_when: ( env.network_mode | upper == 'NAT' ) and (env.jumphost.name is none or env.jumphost.ip is none or env.jumphost.user is none or env.jumphost.pass
```

docs/set-variables-group-vars.md
```diff
@@ -167,4 +167,10 @@
 **env.timezone** | Which timezone would you like Red Hat Enterprise Linux to use? A list of available timezone<br /> options can be found [here](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones). | America/New_York
 **env.ansible_key_name** | (Optional) Name of the SSH key that Ansible will use to connect to hosts. | ansible-ocpz
 **env.ocp_key_name** | Comment to describe the SSH key used for OCP. Arbitrary value. | OCPZ-01 key
-**env.bridge_name** | (Optional) Name of the macvtap bridge that will be created on the KVM host. | macvtap-net
+**env.bridge_name** | (Optional) Name of the macvtap bridge that will be created on the KVM host or in case of NAT the name of the NAT network defenition (usually it is 'default'). If NAT is being used and a jumphost is needed, the parameters network_mode, jumphost.name, jumphost.user and jumphost.pass must be specified, too. In case of default (NAT) network verify that the configured IP ranges does not interfere with the IPs defined for the controle and compute nodes. Modify the default network (dhcp range setting) to prevent issues with VMs using dhcp and OCP nodes having fixed IPs.| macvtap-net
```
jacobemery, line 170: This variable's name should be changed to reflect its more universal nature now that it can account for NAT or macvtap bridge. Since it's not referred to as a NAT bridge and that might be confusing.
Maybe `env.network_name` instead? Make sure to change it here and any time it is referenced in the playbooks, do a global search for 'env.bridge_name' and change all instances to 'env.network_name'.

playbooks/3_setup_kvm_host.yaml
```diff
@@ -86,4 +86,4 @@
   become: true
   roles:
     - configure_storage
-    - macvtap
\ No newline at end of file
+    - { role: macvtap, when: env.network_mode != 'NAT' }
```
jacobemery, line 89:
```suggestion
    - { role: macvtap, when: env.network_mode | upper != 'NAT' }
```

jacobemery: ensure that the task won't fail if users set env.network_mode to lowercase 'nat' instead of 'NAT'

docs/set-variables-group-vars.md
AmadeusPodvratnik: Yes, this makes sense. Due to the output of virsh the name network_name fits. If it's ok I would like to make the global change in a separate pull request.

AmadeusPodvratnik:
Hi @jacobemery I just add the the set_inventory file.

jacobemery:
Doing a quick test and then will merge

jacobemery:
@AmadeusPodvratnik In cases where the jumphost is not the KVM host, an SHH key would need to be copied before the ssh_add_config role could be run on the jumphost.

Could you please add a ssh_copy_id role call to 5_setup_bastion before the ssh_add_config role? You can either make it conditional on whether env.jumphost.ip matches any of the kvm hosts' ips, or just run it without a conditional and if the key has already been copied it will skip it without error, but at least you'll be sure you can connect.

jacobemery:
The network parameter in virt-install looks like it's different when using macvtap versus when using nat. I'm having some trouble with my bastion's networking at the moment, so I found [this](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/virtualization_deployment_and_administration_guide/sect-guest_virtual_machine_installation_overview-creating_guests_with_virt_install) while troubleshooting. The networking issue is not resolved, so I'm not 100% sure this is correct (although I am getting a different error than before because of this change). Please confirm @AmadeusPodvratnik 

NAT: "--network default"
macvtap: "--network network=macvtap"

So the bastion's virt-install command's network parameter should look like this:
"--network {{ 'network=' if ( env.network_mode | upper != 'NAT') else '' }}{{ env.bridge_name }}"

AmadeusPodvratnik:
Hi @jacobemery about the default network in case of virt-install I'm not sure because I found example (check this: https://docs.openstack.org/de/image-guide/virt-install.html or https://linux.die.net/man/1/virt-install see section for network configuration) where the network parameter for default is the same like for others. In case of "network=name" the name of the network configuration on the KVM host will be taken. Due to the fact that I'm not getting an error using --network network=default (this was tested several times) it's hard to say what's going on ... is it possible to troubleshot this on your environment. 
Beside that this opens a good discussion in general. If there is already a macvtap interface on an existing KVM host but 3_setup_kvm_host will be executed (due to additional package install or any other additional requirements we might have) and might fail. So it might make sense to have a flag use existing network or create a new network.

AmadeusPodvratnik:
I added the ssh_copy_id to the setup_bastion playbook to ensure that on a jumphost a ssh key is available.

jacobemery:
> a flag use existing network or create a new network

That would be good, to avoid breaking their existing network.

AmadeusPodvratnik:
> > a flag use existing network or create a new network
> 
> That would be good, to avoid breaking their existing network.

I would suggest to do it in a separate pull request as some sort of '
new feature'

roles/set_inventory/tasks/main.yaml
```diff
@@ -13,12 +13,13 @@
...
   register: find_project
 
-- name: Template out inventory with localhost, FTP, KVM host, and bastion information
+- name: Template out inventory with localhost, FTP, KVM host, jumphost(optional) and bastion information
   tags: set_inventory
   template:
     src: hosts.j2
     dest: "{{ find_project.stdout }}{{ find_inventory.stdout }}/hosts"
     force: yes
+  failed_when: ( env.network_mode | upper == 'NAT' ) and (env.jumphost.name is none or env.jumphost.ip is none or env.jumphost.user is none or env.jumphost.pass is none)
```
jacobemery, line 22:
```suggestion
```

jacobemery: After testing, I realized that this causes other problems. This failed_when condition makes it so that if the task would've failed for other reasons, it doesn't stop the playbook and give an error message.

This is a problem because if one of the other variables used in the inventory are undefined, the user should know it right away.

My apologies!

roles/set_inventory/tasks/main.yaml
```diff
@@ -13,12 +13,13 @@
     echo "${ansible_config%/*}/"
   register: find_project
```
jacobemery, line 15:
```suggestion

- name: Fail if network_mode is NAT and jumphost vars are undefined.
  tags: set_inventory
  fail:
    msg: "Error jumphost vars undefined: when env.network_mode is NAT, you must set all env.jumphost variables."
  when: ( env.network_mode | upper == 'NAT' ) and (env.jumphost.name is none or env.jumphost.ip is none or env.jumphost.user is none or env.jumphost.pass is none or env.jumphost.path_to_keypair is none)

```

jacobemery: Related to comment below... this is the fix for it.

Move the failure condition to its own task.

roles/ssh_cp_keys_jumphost/tasks/main.yaml
```diff
@@ -0,0 +1,19 @@
...
+    block: |
+      Host bastion-jump
+        HostName {{ env.jumphost.name }}
+        User {{ env.jumphost.user }}
+        IdentityFile ~/.ssh/id_rsa
+      Host {{ env.bastion.networking.ip }}
+        HostName {{ env.bastion.networking.ip }} 
+        User {{ env.bastion.access.user }}
+        IdentityFile ~/.ssh/{{ env.ansible_key_name }}
+        ProxyJump bastion-jump
```
jacobemery, line 19: This doesn't need to be its own role because it is only used once, it doesn't require any associated files (like templates), and it's only one task.

Could you please move this task to the end of 0_setup.yaml at the very end of the playbook in a post_tasks section? Thanks!

Like this:

- hosts: localhost
  connection: local
  become: False
  gather_facts: True
  vars_files:
    - "{{ inventory_dir }}/group_vars/all.yaml"
  vars:
    packages: "{{ env.pkgs.controller }}"
    ssh_target: [ "{{ env.ftp.ip }}", "{{ env.ftp.user }}", "{{ env.ftp.pass }}" ]
  roles:
    - install_packages
    - ssh_key_gen
    - ssh_agent
    - ssh_copy_id #to FTP server
  post_tasks:
    - name: Create ssh config file if network mode is NAT
      tags: ssh_copy_id, ssh
      ansible.builtin.blockinfile:
        path: ~/.ssh/config
        backup: true
        create: true
        mode: '0644'
        block: |
          Host bastion-jump
            HostName {{ env.jumphost.name }}
            User {{ env.jumphost.user }}
            IdentityFile ~/.ssh/id_rsa
          Host {{ env.bastion.networking.ip }}
            HostName {{ env.bastion.networking.ip }} 
            User {{ env.bastion.access.user }}
            IdentityFile ~/.ssh/{{ env.ansible_key_name }}
            ProxyJump bastion-jump

docs/prerequisites.md
```diff
@@ -55,7 +55,12 @@ ansible-galaxy collection install ibm.ibm_zhmc
...
 ```
-  * and [Xcode](https://apps.apple.com/us/app/xcode/id497799835?mt=12):
+  * [Xcode](https://apps.apple.com/us/app/xcode/id497799835?mt=12):
 ```
 xcode-select --install
-```
\ No newline at end of file
+```
+## Jumphost for NAT network
+* If for KVM network NAT is used, instead of macvtap, a ssh tunnel using a jumphost is required to access the OCP cluster.  To configure the ssh tunnel expect is required on the jumphost. Expect can be installed by adding it to the Section 11 of the all.yaml file if the jumphost is equal to the KVMi host. Otherwise install it manually on the jumphost:
```
jacobemery, line 63:
```suggestion
* If for KVM network NAT is used, instead of macvtap, a ssh tunnel using a jumphost is required to access the OCP cluster.  To configure the ssh tunnel expect is required on the jumphost. Expect can be installed by adding it to the Section 11 of the all.yaml file if the jumphost is equal to the KVMi host. Otherwise install it manually on the jumphost:
```

jacobemery: This should be automated with a conditional instead of adding a pre-req, if possible.

roles/ssh_add_config/tasks/main.yaml
```diff
@@ -0,0 +1,28 @@
+---
```
jacobemery, line 1: This role no longer adds ssh config and should be renamed.

Forgive me if I'm misunderstanding, what does the ssh_copy_id role lack in capability that prevents it from being re-used here?

docs/set-variables-group-vars.md
```diff
@@ -159,4 +159,10 @@
...
+**env.network_mode: | (Optional) In case the network mode will be NAT and the installation will be executed from remote (e.g. your laptop), a jumphost needs to be defined to let the installation access the bastion host. If macvtap for networking is being used this variable should be empty.
+**env.jumphost.name: | (Optional) If env.network.mode is set to 'NAT' the name of the jumphost (e.g. the name of KVM host if used as jumphost) should be specified.
+**env.jumphost.ip: | (Optional) The ip of the jumphost.
+**env.jumphost.user: | (Optional) The user name to login to the jumphost.
+**env.jumphost.pass: | (Optional) The password for user to login to the jumphost.
+**env.jumphost.path_to_keypair: | (Optional) The path to the keypair on the jumphost to be copied to the bastion.
```
jacobemery, line 168:
```suggestion
**env.jumphost.path_to_keypair: | (Optional) The absolute path to the public key file on the jumphost to be copied to the bastion.
```

roles/ssh_add_config/tasks/main.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+- name: Load in variables
+  tags: ssh_copy_id, ssh
+  include_vars: "{{inventory_dir}}/group_vars/all.yaml"
+
+- name: Use template file to create expect script
+  tags: ssh_copy_id, ssh
+  template:
+    src: ssh-copy-id.exp.j2
```
jacobemery, line 10:
```suggestion
    src: ssh_copy_id.exp.j2
```

roles/ssh_add_config/templates/ssh_copy_id.exp.j2
```diff
@@ -0,0 +1,19 @@
+#!/usr/local/bin/expect -f
```
jacobemery, line 1: Change the name of this file to `ssh-copy-id.exp.j2` to match what is in roles/ssh_add_config/tasks/main.yaml, otherwise the task "Use template file to create expect script" throws an error saying it can't find the file.

roles/ssh_add_config/templates/ssh_copy_id.exp.j2
```diff
@@ -0,0 +1,19 @@
...
+}
+
+set timeout 20
+spawn ssh-copy-id -f -o StrictHostKeyChecking=no -i {{ ssh_target[3] }} {{ ssh_target[1] }}@{{ ssh_target[0] }}
+expect {
+        "password: " {
+                send -- "{{ ssh_target[2] }}\r"
+                expect eof
+        }
+        "Number of key(s) added:" {}
```
jacobemery, line 19:
```suggestion
        "Number of key(s) added:" {}
}
```

jacobemery: Missing closing brace

roles/ssh_add_config/tasks/main.yaml
```diff
@@ -0,0 +1,28 @@
...
+    dest: "/tmp/ssh-copy-id-expect-pass.exp"
+    force: yes
+
+- name: Copy SSH ID from jumphost to bastion with pre-provided password
+  tags: ssh_copy_id, ssh
+  command: "expect /tmp/ssh-copy-id-expect-pass.exp"
+  register: ssh_copy
+
```
jacobemery, line 18:
```suggestion
- name: Installing required package 'expect'.
  ansible.builtin.package:
    name: expect
    state: latest
    update_cache: yes

- name: Copy SSH ID from jumphost to bastion with pre-provided password
  tags: ssh_copy_id, ssh
  command: "expect /tmp/ssh-copy-id-expect-pass.exp"
  register: ssh_copy

```

jacobemery: Required package should be installed automatically

playbooks/5_setup_bastion.yaml
```diff
@@ -1,5 +1,31 @@
...
+  tags: ssh, ssh_copy_id, section_1
+  become: true
+  gather_facts: false
+  vars_files:
+    - "{{ inventory_dir }}/group_vars/all.yaml"
+  vars:
+    ssh_target: ["{{ env.bastion.networking.ip }}", "{{ env.bastion.access.user }}", "{{ env.bastion.access.pass }}", "{{ env.jumphost.path_to_keypair }}", "{{ env.jumphost.ip }}"]
+  roles:
+    - { role: ssh_add_config, tags: ssh_copy_id, ssh, when: (env.network_mode | upper == "NAT") and ( env.jumphost.ip is not none ) }
+
```
jacobemery, line 28:
```suggestion
    ssh_target: ["{{ env.bastion.networking.ip }}", "{{ env.bastion.access.user }}", "{{ env.bastion.access.pass }}", "{{ env.jumphost.path_to_keypair }}", "{{ env.jumphost.ip }}"]
  pre_tasks:
    - name: Generate an OpenSSH keypair with the default values (4096 bits, RSA)
      become: false
      tags: ssh_key_gen, ssh, section_1
      community.crypto.openssh_keypair:
        path: "~/.ssh/id_rsa"
        passphrase: ""
        regenerate: full_idempotence
  roles:
    - { role: ssh_add_config, tags: ssh_copy_id, ssh, when: (env.network_mode | upper == "NAT") and ( env.jumphost.ip is not none ) }

```

jacobemery: SSH key needs to be generated

roles/ssh_cp_keys_jumphost/tasks/main.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+- name: Create ssh config file if network mode is NAT
+  tags: ssh_copy_id, ssh
+  ansible.builtin.blockinfile:
+    path: ~/.ssh/config
+    backup: true
+    create: true
+    mode: '0644'
+    block: |
+      Host bastion-jump
```
jacobemery, line 11:
```suggestion
      Host {{ env.jumphost.ip }}
```

roles/ssh_cp_keys_jumphost/tasks/main.yaml
jacobemery: Just found another change needed that I updated this comment with:

`Host bastion-jump`
and
`ProxyJump bastion-jump`

These settings were returning "Could not resolve hostname jumphost-kvm: Name or service not known" error during TASK [ssh_add_config : Copy SSH ID from jumphost to bastion with pre-provided password].

Changing this to the jumphost IP fixed it.

Changed to:
`Host {{ env.jumphost.ip }}`
and
`ProxyJump {{ env.jumphost.ip }}`

jacobemery: And HostName to env.jumphost.name instead of env.jumphost.name

roles/ssh_add_config/tasks/main.yaml
AmadeusPodvratnik: I commit your suggestions as a interim solution because I think that we need a section in all.yaml in Section 11 for jumphost. I would suggest to change it later.

roles/ssh_add_config/templates/ssh_copy_id.exp.j2
AmadeusPodvratnik: Thx, this was a copy&paste error .... fix pushed ..

roles/ssh_add_config/tasks/main.yaml
AmadeusPodvratnik: Because the ssh_copy_id is deleting a key I don't want to do
There are more parameters e.g. the parameter for env.jumphost.path_to_keypar is needed. So if we want to use ssh_copy_id for general purpose we need to adapt this task. 
My suggestion is here to do it in a separate request.
If move to a more general ssh_copy_id task is ok so far, a rename is not needed because of the delete.
Nevertheless I modified the ssh_add_config because a delete of the fingerprint out of the know_hosts makes sense.

roles/ssh_cp_keys_jumphost/tasks/main.yaml
AmadeusPodvratnik: Unfortunately this might be executed multiple times (every time you create an new bastion this might change because the IP might change - different KVM Lpar, etc). So I do not suggest to move it away from bastion task. Nevertheless I changed the block slightly (not sure if we should deal with IP only because of the readability. The bastion Host parameter I would like to have like bastion.vm_name). Host not found (for jumphost) happens if DNS is not working properly or an /etc/hosts entry is missing but I still switch to ip.

AmadeusPodvratnik:
Hi @jacobemery , I just realized that you add
pre_tasks:
    - name: Generate an OpenSSH keypair with the default values (4096 bits, RSA)
      become: false
      tags: ssh_key_gen, ssh, section_1
      community.crypto.openssh_keypair:
        path: "~/.ssh/id_rsa"
        passphrase: ""
        regenerate: full_idempotence
to the 5_setup_bastion playbook which might destroy an existing id_rsa key (this might lead to other connection issues).  I suggest to remove this part.
The user decides what to do ( which key he would like to use).

AmadeusPodvratnik:
I found a good doc (old and on ubuntu but still valid) how iptables work and what a user need to do afterwards. It describes how port and ip forwarding works. Nevertheless the topic iptables is huge and full of traps and can lead to a disruption of existing connections.
https://www.cyberciti.biz/faq/kvm-forward-ports-to-guests-vm-with-ufw-on-linux/

So at the end I would not suggest to add iptables and let the user add the correct rules.

AmadeusPodvratnik:
@jacobemery I added a global KVM flag (env.z.ip_forward) to all.yaml and add a section to 3_setup_kvm to set the flag (Only for NAT and set it to the value the user specified). Docs are updated, too.

jacobemery (commented):
This variable is currently asking for the public key, so you either have to filter that out like this:
`env.jumphost.path_to_keypair.split('.')[:-1] | join('.')`

playbooks/5_setup_bastion.yaml
```diff
@@ -1,5 +1,39 @@
...
+  vars_files:
+    - "{{ inventory_dir }}/group_vars/all.yaml"
+  vars:
+    ssh_target: ["{{ env.bastion.networking.ip }}", "{{ env.bastion.access.user }}", "{{ env.bastion.access.pass }}", "{{ env.jumphost.path_to_keypair }}", "{{ env.jumphost.ip }}"]
+  pre_tasks:
+    - name: Generate an OpenSSH keypair with the default values (4096 bits, RSA)
+      become: false
+      tags: ssh_key_gen, ssh, section_1
+      community.crypto.openssh_keypair:
+        path: "{{ env.jumphost.path_to_keypair }}"
```
jacobemery, line 31:
```suggestion
        path: "{{ env.jumphost.path_to_keypair.split('.')[:-1] | join('.') }}"
```
The variable is asking for the public key, so the .pub needs to be split off

docs/set-variables-group-vars.md
```diff
@@ -16,6 +16,7 @@
 **Variable Name** | **Description** | **Example**
 :--- | :--- | :---
 **env.z.high_availability** | Is this cluster spread across three LPARs? If yes, mark True. If not (just in<br /> one LPAR), mark False | True
+**env.z.p_forward** | This variable specifies if ip forwarding is enabled or not if NAT network is selected. If ip_forwarding is set to 0, the installed OCP cluster will not be able to access external services. This setting will be configured during 3_setup_kvm playbook. If NAT will be configured after 3_setup_kvm playbook, the setup needs to be done manually before bastion is being created, configured oor reconfigured by running the 3_setup_kvm playbook with parameter: --tags cfg_ip_forward | 1
```
jacobemery:
```suggestion
**env.z.p_forward** | This variable specifies if ip forwarding is enabled or not if NAT network is selected. If ip_forwarding is set to 0, the installed OCP cluster will not be able to access external services. This setting will be configured during 3_setup_kvm playbook. If NAT will be configured after 3_setup_kvm playbook, the setup needs to be done manually before bastion is being created, configured or reconfigured by running the 3_setup_kvm playbook with parameter: --tags cfg_ip_forward | 1
```

jacobemery: Great explanation! Just fixed a typo

docs/set-variables-group-vars.md
```diff
@@ -16,7 +16,7 @@
...
 :--- | :--- | :---
 **env.z.high_availability** | Is this cluster spread across three LPARs? If yes, mark True. If not (just in<br /> one LPAR), mark False | True
-**env.z.p_forward** | This variable specifies if ip forwarding is enabled or not if NAT network is selected. If ip_forwarding is set to 0, the installed OCP cluster will not be able to access external services. This setting will be configured during 3_setup_kvm playbook. If NAT will be configured after 3_setup_kvm playbook, the setup needs to be done manually before bastion is being created, configured oor reconfigured by running the 3_setup_kvm playbook with parameter: --tags cfg_ip_forward | 1
+**env.z.ip_forward** | This variable specifies if ip forwarding is enabled or not if NAT network is selected. If ip_forwarding is set to 0, the installed OCP cluster will not be able to access external services. This setting will be configured during 3_setup_kvm playbook. If NAT will be configured after 3_setup_kvm playbook, the setup needs to be done manually before bastion is being created, configured oor reconfigured by running the 3_setup_kvm playbook with parameter: --tags cfg_ip_forward | 1
```
jacobemery, line 19:
```suggestion
**env.z.p_forward** | This variable specifies if ip forwarding is enabled or not if NAT network is selected. If ip_forwarding is set to 0, the installed OCP cluster will not be able to access external services. This setting will be configured during 3_setup_kvm playbook. If NAT will be configured after 3_setup_kvm playbook, the setup needs to be done manually before bastion is being created, configured or reconfigured by running the 3_setup_kvm playbook with parameter: --tags cfg_ip_forward | 1
```

jacobemery: Great explanation! Just fixed a typo

playbooks/3_setup_kvm_host.yaml
```diff
@@ -81,6 +81,19 @@
...
+  gather_facts: true
+  become: true
+    vars_files:
+      - "{{ inventory_dir }}/group_vars/all.yaml"
+    tasks:
+      - name: Configure ip_forward in case of network "NAT"
+        tags: cfg_ip_forward
+        ansible.builtin.command: sysctl -w net.ipv4.ip_forward={{ env.z.ip_forward }}
+      when: env.network_mode | upper == 'NAT' 
+
```
jacobemery, line 96:
```suggestion
- name: Configure ip_forward in case of NAT
  hosts: kvm_host
  tags: cfg_ip_forward
  gather_facts: true
  become: true
  vars_files:
    - "{{ inventory_dir }}/group_vars/all.yaml"
  tasks:
    - name: Configure ip_forward in case of network "NAT"
      tags: cfg_ip_forward
      ansible.builtin.command: sysctl -w net.ipv4.ip_forward={{ env.z.ip_forward }}
      when: env.network_mode | upper == 'NAT' 
```

jacobemery: Fixed YAML syntax errors

jacobemery: Ah! Another suggestion, because I just tested this and it didn't work because it requires a network reset or a reboot. And I found there's an ansible module for changing sysctl settings! It even includes a reload parameter. Very handy!

jacobemery: Hm, even with the reload, this isn't working unless I disable firewalld. Going to have to keep investigating...

playbooks/3_setup_kvm_host.yaml
```diff
@@ -81,9 +81,26 @@
...
+  tasks:
+    - name: Configure ip_forward in case of network "NAT"
+       tags: cfg_ip_forward
+       ansible.posix.sysctl:
+         name: net.ipv4.ip_forward
+         value: "{{ env.z.ip_forward }}"
+         sysctl_set: true
+         state: present
+         reload: true
+       when: env.network_mode | upper == 'NAT'
```
jacobemery, line 100:
```suggestion
    - name: Configure ip_forward in case of network "NAT"
      tags: cfg_ip_forward
      ansible.posix.sysctl:
        name: net.ipv4.ip_forward
        value: "{{ env.z.ip_forward }}"
        sysctl_set: true
        state: present
        reload: true
      when: env.network_mode | upper == 'NAT' 
```

jacobemery: I made an indentation error when copying this over, apologies!

docs/prerequisites.md
```diff
@@ -55,7 +55,12 @@ ansible-galaxy collection install ibm.ibm_zhmc
...
 ```
 xcode-select --install
-```
\ No newline at end of file
+```
+## Jumphost for NAT network
+* If for KVM network NAT is used, instead of macvtap, a ssh tunnel using a jumphost is required to access the OCP cluster.  To configure the ssh tunnel expect is required on the jumphost. Expect can be installed by adding it to the Section 11 of the all.yaml file if the jumphost is equal to the KVM host. Otherwise install it manually on the jumphost:
+```
+yum install expect 
+```
```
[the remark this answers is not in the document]
AmadeusPodvratnik: Change doc and update description.

smolin-de (commented):
Install of "python3" on a RHEL8 host will install python 3.6.
This is an unsupported python version!
I am installing and using  "python3.9" package on my RHEL8 based LPARs

I recommend not to install python on a RHEL8 host. It is better to report an error, and let the user decide which version he wants to install manually. 
This is a one-time setup !  Therefore, it can be executed manually.

We do not have such issue on RHEL9.  
On a RHEL9 KVM host, I would agree to install the package automatically.

smolin-de (commented):
Install of "python3" on a RHEL8 host will install python 3.6.
This is an unsupported python version!
I am installing and using "python3.9" package on my RHEL8 based LPARs

We do not have such issue on RHEL9.
On a RHEL9 KVM host, I would agree to install the package automatically.

jacobemery:
Ok! I think this is ready to go!! I tested out a full run through with NAT and macvtap to make sure they both work. Both full, clean run-throughs. Thanks for all the work on this @AmadeusPodvratnik !!

PR #78, opened 2023-01-24
feat: Improve OCP versioning

jacobemery:
Previous method for selecting the OpenShift version in all.yaml was unnecessarily cumbersome, requiring copying 5 different mirror links to variables.

This make it so you just set one variable env.openshift.version to the desired version (i.e. 4.12) and the automation does the rest.

### DIFF
--- a/roles/get_ocp/tasks/main.yaml
+++ b/roles/get_ocp/tasks/main.yaml
@@ -37,7 +37,7 @@
   get_url:
     url: "https://mirror.openshift.com/pub/openshift-v{{ env.openshift.version | string | split('.') | first }}/{{ env.install_config.control.architecture }}/dependencies/rhcos/{{ ( env.openshift.version | string | split('.') )[:-1] | join('.') }}/{{ env.openshift.version }}/rhcos-live-rootfs.{{ env.install_config.control.architecture }}.img"
     dest: "/var/www/html/bin/rhcos-live-rootfs-{{ env.openshift.version }}-{{ env.install_config.control.architecture }}.img"
-    mode: '0755'
+    mode: '0644'
 
 - name: Unzip OCP client and installer
   tags: get_ocp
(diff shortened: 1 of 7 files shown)


jacobemery:
@smolin-de Could you review this for me?

jacobemery (commented):
References to architecture and references to the major version in the mirror links have also been turned into variables for future-proofing.

smolin-de:
In general I like this change.
You are using just "major.minor" version (4.12) and getting always the the latest fix version. So you always install the latest available version.
Was this requested from someone ?
I prefer to specify the entire version, e.g. 4.12.0. I can always updated to the latest version later, if and when I like.
This give me more flexibility.

jacobemery:
@smolin-de when merging, could you change this PRs title to "feat: improve OCP versioning #78" to match the semantic release naming conventions?

roles/get_ocp/tasks/main.yaml
```diff
@@ -1,16 +1,13 @@
 ---
 
-- name: Delete bin and ignition folders when refresh_ocp_version has been passed as True to command-line via extra-vars.
-  tags: get_ocp, refresh_ocp_version
-  vars:
-    refresh_ocp_version: 'False'
+- name: Delete bin and ignition folders for idempotency.
```
smolin-de, line 3: In future,  we should only delete files which could cause problems.

smolin-de (changes requested):

roles/get_ocp/tasks/main.yaml
jacobemery: https://github.com/IBM/Ansible-OpenShift-Provisioning/pull/78/commits/cde53950ce71a4d0dfec9a530812ecf460dab5a1

smolin-de (approved):

PR #81, opened 2023-01-24
Additional artifact refresh

jacobemery:
Add prep_kvm_guests role call to reinstall_cluster playbook so that new versions of the kernel and initramfs are pulled down if the OCP version changes in inventory.

### DIFF
--- a/playbooks/reinstall_cluster.yaml
+++ b/playbooks/reinstall_cluster.yaml
@@ -1,6 +1,6 @@
 # Use this if you want to re-install nodes with a new OCP version
 ---
-- name: Refresh initramfs and kernel.
+- name: Update initramfs and kernel.
   hosts: kvm_host
   become: True
   gather_facts: False
@@ -9,7 +9,7 @@
   roles:
     - prep_kvm_guests
 
-- name: Refresh ignitions and other install files.
+- name: Update ignitions and other install files.
   hosts: bastion
   become: True
   vars_files: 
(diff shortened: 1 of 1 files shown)


jacobemery:
@smolin-de please review this as well.

playbooks/reinstall_cluster.yaml
```diff
@@ -1,7 +1,16 @@
 # Use this if you want to re-install nodes with a new OCP version
 ---
+- name: Refresh initramfs and kernel.
```
smolin-de, line 3: `- name: Update initramfs and kernel`

playbooks/reinstall_cluster.yaml
```diff
@@ -1,7 +1,16 @@
...
+  hosts: kvm_host
+  become: True
+  gather_facts: False
+  vars_files:
+    - "{{ inventory_dir }}/group_vars/all.yaml"
+  roles:
+    - prep_kvm_guests
 
-- hosts: bastion
+- name: Refresh ignitions and other install files.
```
smolin-de, line 12: To have a play name is good, but I do not agree with the text.
I doubt, that you are refreshing ignition files. I did not saw such code yet.
You create them.
use: "Create ignitions and other install files"

jacobemery: > To have a play name is good, but I do not agree with the text. I doubt, that you are refreshing ignition files. I did not saw such code yet. You create them. use: "Create ignitions and other install files"

Line 40 of get_ocp role:

`- name: Delete OCP download directory for idempotency, because ignition files deprecate after 24 hours.
  tags: get_ocp
  become: true
  file:
    path: /root/ocpinst
    state: absent`

playbooks/reinstall_cluster.yaml
jacobemery: Update is a better word.

smolin-de (approved):

PR #82, opened 2023-01-24
fix: Group virt does not exist in RHEL9

smolin-de:
The yum group @virt does not exists anymore in RHEL9. We install now the files which where defined in that group.
This solution will work for RHEL8 and RHEL9 KVM host installs.


jacobemery (approved):
lgtm

PR #86, opened 2023-01-28
feat: Automate install of zhmc dependencies

jacobemery:
Automatically install ibm.ibm_zhmc collection's python package dependencies, and update docs to match.


smolin-de:
I suggest to separate the install code from setup/configure code. Except, install bastion. 
So, the install LPAR and install KVM code should be included to a separate tool.

smolin-de (approved):

PR #91, opened 2023-02-01
chore: Change default cluster network to OVNKubernetes

smolin-de:
The default OCP cluster network is OVNKubernetes.


jacobemery (approved):
Always good to align with the install docs, thanks!

PR #94, opened 2023-02-02
fix: Create bootstrap VM on RHEL9 based KVM host

smolin-de:
The virt-install command on RHEL9 KVM hosts requires a os-variant parameter.


jacobemery:
@smolin-de I vaguely remember us talking about this a while back, but I don't remember what conclusion we came to about it - could we change this to a variable instead of hard-coding the value?

smolin-de:
> @smolin-de I vaguely remember us talking about this a while back, but I don't remember what conclusion we came to about it - could we change this to a variable instead of hard-coding the value?

yes, we can have a variable.
I will create a new pull request.

PR #98, opened 2023-02-11
fix: Escape host and group vars from .gitignore

jacobemery:
Escape host and group vars from .gitignore in inventories/default, otherwise changes to the all.yaml.template will need to be forced when committed.


smolin-de (approved):
Good catch

PR #99, opened 2023-02-11
feat: improve delete_nodes role

jacobemery:
Add deletion of the bootstrap node to the delete_nodes role, in case of restarting 6_create_nodes before bootstrap is already deleted.

Add changed_when parameter so that "ok" is given when nodes are not destroyed or undefined, and "changed" when they are.

### DIFF
--- a/roles/delete_nodes/tasks/main.yaml
+++ b/roles/delete_nodes/tasks/main.yaml
@@ -1,5 +1,14 @@
 ---
 
+- name: Delete bootstrap node, if exists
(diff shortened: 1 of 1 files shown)


smolin-de (approved):

smolin-de:
I would change the commit message.
For, me it is not a feature. The deletion of nodes was already there. You just improved the code.
So, I would use:
fix: Improve delete_nodes role 
or
chore: Improve delete_nodes role

Issue #96, opened 2023-02-08
FTP server documentation needs update


jacobemery:
Now that the KVM host serves everything the bastion needs to boot via HTTP, the FTP server and its variables need to be updated to reflect that.

An FTP server is still required when the user is booting the KVM host(s) for the first time, but after that it is not required. 

I haven't figured out yet how to differentiate / describe these two functions in all.yaml and the documentation.

PR #102, opened 2023-02-22
fix: Change ftp server to file server

jacobemery:
Related to Issue #96 

The name 'FTP server' is confusing since the KVM host is now capable of hosting and serving necessary files via http only. Update the docs and variables to represent this change. 

Also, enable using either ftp or http to serve config files for the bastion. FTP is still the only option for booting the KVM hosts, as that is the only option for the HMC to network boot from. And HTTP will still be installed and used on the bastion as it is required. But with these changes, someone using a pre-existing KVM host and NAT-based networking can use only HTTP if they'd like. And someone using an FTP server for booting of the KVM host(s) can use that same FTP server to boot the bastion.

### DIFF
--- a/docs/set-variables-group-vars.md
+++ b/docs/set-variables-group-vars.md
@@ -40,7 +40,7 @@
 **env.file_server.user** | Username to connect to the file server. Must have sudo and SSH access. | user1
 **env.file_server.pass** | Password to connect to the file server as above user. | user1pa$s!
 **env.file_server.protocol** | Protocol used to serve the files, either 'ftp' or 'http' | http
-**env.file_server.iso_mount_dir** | Directory path relative to the HTTP/FTP accessible directory where RHEL ISO is mounted. For example, if the FTP root is at /home/user1<br /> and the ISO is mounted at /home/user1/RHEL/8.5 then this variable would be<br /> RHEL/8.5 - no slash before or after. | RHEL/8.5
+**env.file_server.iso_mount_dir** | Directory path relative to the HTTP/FTP accessible directory where RHEL ISO is mounted. For example, if the FTP root is at /home/user1<br /> and the ISO is mounted at /home/user1/RHEL/8.7 then this variable would be<br /> RHEL/8.7 - no slash before or after. | RHEL/8.7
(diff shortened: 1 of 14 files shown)


jacobemery:
Received feedback from users that this was confusing to them, decided to tackle it today. Tested a highly-available deployment (3 LPARs, macvtap) with env.file_server.protocol as 'ftp' and entire test run as 'http'.

smolin-de (changes requested):
This statement:
"KVM host is now capable of hosting and serving necessary files via http"
is WRONG !
A guest on a KVM host does not have a network connection to the 
host via macvtap !

docs/set-variables-group-vars.md
```diff
@@ -33,14 +33,15 @@
...
-**env.ftp.cfgs_dir** | Directory path relative to FTP root where configuration files can be stored. If FTP root is /var/ftp/pub<br /> and you would like to store the configs at /var/ftp/pub/ocpz-config then this variable would be<br /> ocpz-config. No slash before or after. | ocpz-config
+**env.file_server.ip** | IPv4 address for the file server that will be used to pass config files and<br /> iso to KVM host LPAR(s) and bastion VM during their first boot. | 192.168.10.201
+**env.file_server.user** | Username to connect to the file server. Must have sudo and SSH access. | user1
+**env.file_server.pass** | Password to connect to the file server as above user. | user1pa$s!
+**env.file_server.protocol** | Protocol used to serve the files, either 'ftp' or 'http' | http
+**env.file_server.iso_mount_dir** | Directory path relative to the HTTP/FTP accessible directory where RHEL ISO is mounted. For example, if the FTP root is at /home/user1<br /> and the ISO is mounted at /home/user1/RHEL/8.5 then this variable would be<br /> RHEL/8.5 - no slash before or after. | RHEL/8.5
```
smolin-de, line 43: **env.file_server.iso_mount_dir** | Directory path relative to the HTTP/FTP accessible directory where RHEL ISO is mounted. For example, if the FTP root is at /home/user1<br /> and the ISO is mounted at /home/user1/RHEL/8.7 then this variable would be<br /> RHEL/8.7 - no slash before or after. | RHEL/8.7

jacobemery: Done

jacobemery:
> A guest on a KVM host does not have a network connection to the KVM host via macvtap !
Ah, you are correct, my apologies! I've updated the description. Looks like none of the code has to change, but that was a lapse in my understanding, thanks!

Everything else look ok @smolin-de ?

smolin-de (approved):
ok

Issue #103, opened 2023-02-22
Create optional playbook to setup file server


jacobemery:
It would be nice for users, and relatively simple, to create a playbook that sets up the file server in a way that is compatible with these playbooks.

If the file server is going to be used to boot KVM hosts, it must have FTP, otherwise only HTTP is required to be installed and configured, unless otherwise specified by the env.file_server.protocol variable.

PR #105, opened 2023-02-23
feat: Add optional setup_file_server.yaml playbook

jacobemery:
Related to issue 'Create optional playbook to setup file server' #103.

Add a playbook to automatically setup the file server. Based on feedback from users, the file server's setup causes a lot of confusion. 

This playbook is not part of the official install path, but meant to be a helpful additional resource if needed. It would serve as an "executable documentation" for setting up the file server. Either this works as-is for some user's needs; or for those that it doesn't meet their needs, it at least lays out precisely what must be done.

Also adds accompanying documentation in prerequisites and run_the_playbooks pages.


jacobemery:
@smolin-de what do you think?

jacobemery:
Ah thanks for the comment @smolin-de. I didn't see it until now because it looks like you edited my comment with your response. I didn't even know that could be done! 

I think it's helpful for up setting up new environments, and as a form of "executable documentation" where all the setup required is laid out plainly, and can be run if needed.

Does anything need to be modified in your opinion, or do you have any other thoughts? Still needs your official review.

AmadeusPodvratnik:
@jacobemery I went through the changes and have some comments:
- It's not clear where to install the FTP server. Basically a FTP server can be installed on KVM but than macvtap is not working. Using NAT, nodes have access to the KVM host. Make it sense to mention it. 
- Make it sense to have a simple index page user can query (e.g. curl command). E.g.:
<html>
<head/>
<body>Installation of FTP Server was successful!/body>
</html>
- Why you removed the mount commands from documentation? I feel they are helpful and worth to be mentioned.

AmadeusPodvratnik (changes requested):
Thx @jacobemery for the PR. I tested it and have following comments:
- due to the fact that a file_server entry in the hosts file is needed, this should be documented.
- Make it sense to automate this step in case an user would like to have it "all together". E.g. if a user would like to have it all self contained than this step can be integrated into a more general setup playbook. 
- For macvtap network mode any other VM can be used if the LPAR has access to. Is this something you would like to mention?

jacobemery:
Needs rebase.

PR #109, opened 2023-04-07
fix: Remove '3_setup_kvm_host' from 'pre-existing_site' playbook

smolin-de:
Sudo access is required to setup a KVM host. In a pre-existing env, we assume that the KVM host is already installed.

### DIFF
--- a/playbooks/pre-existing_site.yaml
+++ b/playbooks/pre-existing_site.yaml
@@ -2,7 +2,6 @@
 ---
 
 - import_playbook: 0_setup.yaml
-- import_playbook: 3_setup_kvm_host.yaml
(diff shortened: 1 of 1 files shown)


jacobemery (approved):
Originally I was thinking that this playbook would be used when an LPAR had been created and RHEL pre-installed, but not necessarily set-up to be a KVM host yet. So that it would be helpful for people who wanted to just set up a RHEL server and have this playbook do the rest. 

I see what you're saying though, especially now that we've enabled running these playbooks as a user w/o sudo access.

Either way the user can easily add/remove import_playbook statements from here as needed. I like the idea how isolating sudo access with this playbook though.

PR #111, opened 2023-04-08
fix: Add timeout for bastion and bootstrap VM creation

smolin-de:
Creation of bastion or bootstrap VM's sometimes hang forever. This patch set a timeout of 5 min for VM creation.


jacobemery (approved):
Good idea, looks good to me

PR #116, opened 2023-04-20
chore: Add common role to support future multi arch setups

smolin-de:
Add a common role tu support s390x and x86 KVM systems in furture. Most of the functions and variables are not used yet.


smolin-de:
@jacobemery 
can you please review and approve my PR

AmadeusPodvratnik (commented):
Hi @smolin-de,
quick question here: Looking at your code I did not see an existing console statement in prior releases, correct?
Using Assisted Installer (s390x KVM) the console statement (additional kernel args) for the virt-install command is "console=tty1 console=ttyS1,115200n8". 
ttysclp0 is used for zVM nodes (parm line). Maybe the AI example for KVM nodes might work as a reference? How do you know that ttysclp0 is the correct value for KVM?
With the integration of zVM this var might change ...

smolin-de:
> Hi @smolin-de, quick question here: Looking at your code I did not see an existing console statement in prior releases, correct? Using Assisted Installer (s390x KVM) the console statement (additional kernel args) for the virt-install command is "console=tty1 console=ttyS1,115200n8". ttysclp0 is used for zVM nodes (parm line). Maybe the AI example for KVM nodes might work as a reference? How do you know that ttysclp0 is the correct value for KVM? With the integration of zVM this var might change ...

Q - How do you know that ttysclp0 is the correct value for KVM? 
A - It is documented here: 
https://www.ibm.com/docs/en/linux-on-systems?topic=access-virsh-kvm-guest . I tried it out and it works. So, I implemented it.

With integration of zVM we may need to change the code to distinguish between zKVM an zVM.

PR #119, opened 2023-05-02
Changes to handle creation of AgentServiceConfig, Hosted Control Plane and InfraEnv for agent based hypershift installation

veera-damisetti:
Installs prerequisites on kvm host and will Create AgentServiceConfig, Hosted Control Plane and InfraEnv for the
agent based hypershift installation


AmadeusPodvratnik (changes requested):
In roles/create_agentserviceconfig_hypershift/tasks/main.yaml you have a hard coded curl command to mirror and there on 4.13.0-ec.4 which should be depended on the installed OCP version.
For all.yaml.template you should not hardcoded the id_rsa.pub keys but describe it in the documentation for the new parameters.
PULL_SECRET is already present so you can reuse it (or is this something different)?
In roles/install_prerequisites_host_hypershift/tasks/main.yaml there is a hardcoded id_rsa key. Pls use instead parameters or for example the ansible ssh key.
I do not see the description of the newly created parameters. So pls add the description to the doc as part of this PR (This helps to understand the code much better)

AmadeusPodvratnik (changes requested):
Sorry for not being precise enough where to put the documentation for the parameters to. Pls add the documentation of the vars not in the all.yaml.template but here: (https://github.com/IBM/Ansible-OpenShift-Provisioning/tree/main/docs)/set-variables-group-vars.md.
In this docs folder you find the documentation of the whole project. So pls add the variable documentation and the documentation of the playbooks and how to use them to. 
Explaining e.g. how you handle the pull secret is necessary for the overall  understanding how to setup the variables. In addition you need to describe the playbooks and how to use them. I believe we need a group for hypershift (@jacobemery ) or a new number for the playbooks and / or a specific section in the documentation. If hypershift will be used the user should know what to do. E.g. install a OCP cluster (link to the documentation) and than run playbook 1 and 2 .... etc. 
You are using a mirror which is necessary in case of stage or iteration systems in RH. If production for RH is used I believe there is no mirror necessary, correct? Maybe you introduce a "dev" flag to interact with RH mirrors?
For the AgentServiceConfig is the OCP version for AgentServiceConfig decoupled from the installed OCP version or is it bound to it?
Installing prereqs on the KVM host is a root task and not sure if your approach is working for not root access. Did you tested it with none root user? If the KVM host need some additional packages it might be good to add software installations in the setup_kvm_host playbook. Did you need all of the packages? E.g. wget where I did not found a execution statement. On KVM I believe that you have already libvirt running so pls check install_prerequisites_host_hypershift role and check if thinks can be put into setup_kvm_host playbook.

veera-damisetti:
We can put prereqs  into   setup_kvm_host playbook but i beleive , having separate playbook for hypershift which is independent of creation of managemen cluster  is better .
which means , if user has management cluster already and trying to install hypershift, he can use hypershift playbooks only 
instead of installing all prereqs required for management cluster .
same thing occurs if user want second hosted cluster on single management cluster from different host.

AmadeusPodvratnik:
@veera-damisetti 
 What of those packages:
      - wget
      - make
      - jq
      - qemu-kvm
      - libvirt
      - libvirt-devel
      - libvirt-daemon-kvm
      - qemu-img
      - genisoimage
      - virt-manager 
      - "@Development Tools"
      - git
      - virt-install
 are required? Some of them are already part of kvm installation. What packages are specific to hypershift? Beside a specific role you can use a tags and describe how to execute a playbook with your tag to just setup the hypershift depended packages.

veera-damisetti:
yeah these package installations i can move to existing role install_packages and will call with hypershift tag .
But hypershift specific like , go , hypershift binary and logging in to management cluster, 
these i can keep in my  role right ..?

AmadeusPodvratnik:
Yes, for hypershift related packages you can keep your own role.

veera-damisetti:
For the AgentServiceConfig,  the OCP version for AgentServiceConfig is not bounded to  installed OCP version of Management Cluster. User can have different version for AgentServiceConfig . So giving an option to user to give version for AgentServiceConfig

veera-damisetti:
Updated the Description for variables to [set-variables-group-vars.md]. but how about the documentation for procedure 
Shall i create one separate file for documentation of procedure for Hypershift...?

AmadeusPodvratnik:
@veera-damisetti I would say to create a separate doc handling all relevant steps for hypershift. @jacobemery what Do you think?

AmadeusPodvratnik (changes requested):
In all.yaml we are using lower case variable names but you mix them (lower and upper case) and from a consistency point of view I suggest to make all parameters lower case.

Question: 
- Installing packages is basically a root task. Do you need root access on KVM host? The latest code base has a regression where root access was removed. Did you test hypershift with latest code base?
- It looks like that you are installing the OC client on the KVM host but there should be a matching OC client on the bastion. Maybe you can rollout hypershift nodes from bastion. Can you pls explain this step for better understanding?

veera-damisetti:
Replying for question 2 : We are going to boot the nodes from KVM host . Bastion we are using only  for configuring httpd ( for rootfs image ) and for Configuring HAproxy.

AmadeusPodvratnik (changes requested):
Hi @veera-damisetti , I recognize that you have a description in the .md and a slight different description in the all.yaml.template. First is to say that we should document in the .md files only (there is no need to document vars in the template file). In addition is to say that you should combine the documentation within .md and template files in the .md files.
E.g. kvm_host:
vm_host: #kvm host where you want to run all oc commands and create VMs ( Agents )
**hypershift.kvm_host** | IPv4 address of KVM host for hypershift | 192.168.10.1
Pls remove the documentation from all.yaml.template and combine it together in set-variables-group-vars.md file.
In addition I suggest to move the secrets.yaml into inventory/default folder and generate a template out of it (like all.yaml). This would be consistent to the existing way how we handle our variables (pls adapt the documentation accordingly).

veera-damisetti:
okay Added the necessary description to .md file and removed the description from template file

AmadeusPodvratnik (approved):
Hi @veera-damisetti , thanks for the changes. From my side it's looking good so far.

veera-damisetti:
Thanks .. ! @AmadeusPodvratnik

PR #124, opened 2023-05-04
chore: Disable OS-Info, required for RHEL 9

smolin-de:


smolin-de:
@Jacob,
this patch should not break existing functions. Please merge.
FYI.
@AmadeusPodvratnik

AmadeusPodvratnik (commented):
Will this change have an impact on older but still supported OCP versions (e.g. 4.11 or 4.12)?

smolin-de:
> Will this change have an impact on older but still supported OCP versions (e.g. 4.11 or 4.12)?

No impact to OCP.

PR #126, opened 2023-05-05
fix: Separate RHCOS and OCP versions

smolin-de:
This patch provides the user the ability to specify the OCP and the RHCOS version separately.

### DIFF
--- a/docs/set-variables-group-vars.md
+++ b/docs/set-variables-group-vars.md
@@ -174,7 +174,7 @@
 
 **Variable Name** | **Description** | **Example**
 :--- | :--- | :---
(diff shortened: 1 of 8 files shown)


smolin-de:
FYI
@AmadeusPodvratnik

AmadeusPodvratnik (commented):
@smolin-de even it's somehow cumbersome I would integrate the documentation to this PR. Documentation is not that much and saves another PR only for documentation. Code changes looking good to me.

AmadeusPodvratnik (commented):
@smolin-de Thx for adding the doc to the PR. I would suggest to remove the "Feel free...." because this not the case for other parameters. In addition I would write "Link to the CoreOS files..." and not mention mirror because you can configure any valid URL.

smolin-de:
@AmadeusPodvratnik 
I added the suggested changes.

AmadeusPodvratnik (approved):
Thank you. I'm good with those changes.

PR #139, opened 2023-05-25
fix: Increase csr wait time

smolin-de:
Increase csr wait time, to resolve low network performance issues.


AmadeusPodvratnik (approved):
Code looks good ...

PR #145, opened 2023-05-26
chore: Install bootstrap on first KVM host

smolin-de:
For multi-arch we need to define two KVM hosts. The second host is motlikly based on a different architecture. We need to make sure that the bootstrap host is based on the same architecture as the control node host.


AmadeusPodvratnik (approved):
Maybe it would be good to give the user the opportunity to decide where to put the bootstrap to. This can be implemented in a later step. For the time being let's keep it that simple.

PR #153, opened 2023-06-01
feat: Add and delete compute nodes (day-2)

smolin-de:
Add support to create additional compute nodes (day-2) with different architecture and to delete existing compute nodes.

### DIFF
--- a/docs/run-the-playbooks.md
+++ b/docs/run-the-playbooks.md
@@ -3,10 +3,10 @@
 * Navigate to the [root folder of the cloned Git repository](https://github.com/IBM/Ansible-OpenShift-Provisioning) in your terminal (`ls` should show [ansible.cfg](https://github.com/IBM/Ansible-OpenShift-Provisioning/blob/main/ansible.cfg)).
 * Run this shell command:
 ```
-ansible-playbook playbooks/0_setup.yaml 
+ansible-playbook playbooks/0_setup.yaml
 ```
 
-* Run each part step-by-step by running one playbook at a time, or all at once using [playbooks/site.yaml](https://github.com/IBM/Ansible-OpenShift-Provisioning/blob/main/playbooks/site.yaml). 
+* Run each part step-by-step by running one playbook at a time, or all at once using [playbooks/site.yaml](https://github.com/IBM/Ansible-OpenShift-Provisioning/blob/main/playbooks/site.yaml).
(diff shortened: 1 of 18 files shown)


AmadeusPodvratnik:
> > Looking at the delegate_to command makes me wonder. Did you consider that the hostname is not part of the group file of the inventory and can be any KVM host beside the KVM hosts used for the installation, than it might be good to preconfigure the IP within the network interface and using a mac address instead.
> 
> The user must make sure that ansible has access to the host. Ether define it in the all.yaml file (using 2nd LPAR) or define it manually in the hosts file.
> 
> I do not understand what you mean with using mac address ?
In case of DHCP network definition it might be useful to think about mac addresses instead of IP addresses. Day-2 operations might act differently than day-1. In addition for multiarch it might be that the value for "env.bridge_name" 
is not available or need to specified separately for the new compute nodes running on a KVM host not be in the groups file.

smolin-de:
> > > Looking at the delegate_to command makes me wonder. Did you consider that the hostname is not part of the group file of the inventory and can be any KVM host beside the KVM hosts used for the installation, than it might be good to preconfigure the IP within the network interface and using a mac address instead.
> > 
> > 
> > The user must make sure that ansible has access to the host. Ether define it in the all.yaml file (using 2nd LPAR) or define it manually in the hosts file.
> > I do not understand what you mean with using mac address ?
> > In case of DHCP network definition it might be useful to think about mac addresses instead of IP addresses. Day-2 operations might act differently than day-1. In addition for multiarch it might be that the value for "env.bridge_name"
> > is not available or need to specified separately for the new compute nodes running on a KVM host not be in the groups file.

AmadeusPodvratnik:
> > In case of DHCP network definition it might be useful to think about mac addresses instead of IP addresses. Day-2 operations might act differently than day-1. In addition for multiarch it might be that the value for "env.bridge_name"
> > is not available or need to specified separately for the new compute nodes running on a KVM host not be in the groups file.
> 
> I do not plan to develop DHCP network definition support. This could be an enhancement in future. I am using static IPs to add compute nodes in a day-2 mode since several weeks and it is working just fine.

Ok understood. Than the open point is the env.bridge.name for an additional KVM host. Is it required to have, cross all KVM lpars, the same network definition names (env.bridge.name)?

AmadeusPodvratnik (approved):
Code looks good. Thx.

PR #168, opened 2023-06-16
fix: Check for 'PROGRESSING' state, instead 'AVAILABLE' state

smolin-de:
Some cluster operators are in 'AVAILABLE' state, but are still in 'PROGRESSING' state. We must make sure that all cluster operators have completed their setup.


AmadeusPodvratnik (approved):
Code looks good.

PR #171, opened 2023-06-21
fix: Update Bastion DNS entries only when enabled

smolin-de:
Update Bastion DNS entries only when enabled.

### DIFF
--- a/playbooks/create_compute_node.yaml
+++ b/playbooks/create_compute_node.yaml
@@ -43,6 +43,7 @@
     - role: update_ignition_files
 
     - role: dns_update
+      when: env.bastion.options.dns
       param_dns_cmd: add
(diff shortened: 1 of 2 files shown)


AmadeusPodvratnik (commented):
Hi @smolin-de, I discovered issues with undefined vars. Makes it sens to have a "is defined" check, too?

AmadeusPodvratnik:
> > Hi @smolin-de, I discovered issues with undefined vars. Makes it sens to have a "is defined" check, too?
> 
> Hi, env.bastion.options.dns is a defined variable in the all.yaml.template file. It must be defined. Either 'true' or 'false'. If the variable is not defined, do we assume bastion host supports dns or not ?

I would say that DNS is being supported if set to true all other values (include not define) will be taken as not supported.

jacobemery (approved):
LGTM

PR #175, opened 2023-07-08
feat: Add creation of agents for hypershift

veera-damisetti:
1. Handles creation  of agents for hypershift and scales the nodepool

- Create Agents
- Scale up nodepool
- Monitor the creation of Resources
- wait for installation to complete

2. Handles the deletion of all the resources 

- Destroys Hosted Control Plane
- Deletes all the resources created as part of installation

**Updated docs for the same**


AmadeusPodvratnik (commented):
Code looks good so far. One question @veera-damisetti : Is this a fixed value: until: pod_count.stdout | int > 30 (in roles/create_hcp_InfraEnv_hypershift/tasks/main.yaml). Can this somehow vary from version to version and is there a possibility to do this check differently?
In roles/delete_resources_kvm_host_hypershift/tasks/main.yaml I assume you can add --remove-all-storage to the destroy command so the additional loop can be removed. The same for the bastion.

veera-damisetti:
@AmadeusPodvratnik Thanks for the review 
  pod count might  some how vary from version to version , but after we got this amount of pods we are waiting for all pods to come  up to running state . so this combination can handle the scenario to wait until HCP is in good state

PR #180, opened 2023-07-20
Updated change  to use ansible module for ssh key generation

veera-damisetti:
- Updated change  to use ansible module for ssh key generation
- Giving comment for key as Ansible-OpenShift-Provisioning SSH key


AmadeusPodvratnik (approved):

veera-damisetti:
Thanks @AmadeusPodvratnik

PR #190, opened 2023-08-09
Enhancement: Added support for non root user to run hypershift playbooks

veera-damisetti:
Added support for non root user to run hypershift playbooks

- Updated playbooks to allow users with sudo privileges to run the hypershift playbooks.
- Added change to wait for MCE to be available after creating AgentServiceConfig.

Updated documentation for the same


veera-damisetti:
Thanks @AmadeusPodvratnik

PR #191, opened 2023-08-11
Fix : Changed ssh key path - Hypershift 

veera-damisetti:
Changed ssh key path - Hypershift 

- Changed the path for ssh key on local host 
- Removed space while creating HCP which is causing issue while using different basedomian


AmadeusPodvratnik (approved):
Looks good. Thx.

PR #193, opened 2023-08-16
feat: Detect automatically required RHCOS images for compute nodes

smolin-de:
This patch detects the required RHCOS images to install a compute node.


AmadeusPodvratnik (approved):
As discussed .. good to go.

PR #194, opened 2023-09-01
Enhancement: Changed port for httpd configuration

veera-damisetti:
Changed port for httpd configuration

- Earlier we used port 443 and 8080 for httpd , but we have conflict with port 443 while configuring HAProxy in later stages.
- So changed the port from 443 to 8443


AmadeusPodvratnik (approved):
Looks good so far.

Issue #188, opened 2023-08-03
playbooks/3_setup_kvm_host.yaml failes, when 'ansible_user: root' is used


smolin-rh:
playbooks/3_setup_kvm_host.yaml failes, when 'ansible_user: root' is used.

```
TASK [Set libvirt management to libvirt group instead of root.] *****************************************************************************
ok: [a3elp60]

TASK [Create file for user's custom libvirt configurations.] ********************************************************************************
fatal: [a3elp60]: FAILED! => {"changed": false, "msg": "Error, could not touch target: [Errno 2] No such file or directory: b'/home/root/.config/libvirt/libvirt.conf'", "path": "/home/root/.config/libvirt/libvirt.conf"}
```

AmadeusPodvratnik:
Fix with PR #197

AmadeusPodvratnik:
Code is merged.

PR #197, opened 2023-09-08
fix: root user support - issue #188

AmadeusPodvratnik:
Fix for issue #188 (root user suport). In addition set 0 for ip_forward  var as default because macvtap is default network setting. In addition add var for subscription management.


jacobemery (changes requested):
Since the decision to attach a subscription or not is a yes or no question, could you change env.redhat.manage_subscription to a boolean value? That way you don't have to normalize the case of the string, or reference the docs to see what the available options are. Also if it's just true of false, it's simpler to reference the variable in a playbook, you can just write "when: env.redhat.manage_subscription".

docs/set-variables-group-vars.md
```diff
@@ -48,6 +48,7 @@
 :--- | :--- | :---
 **env.redhat.username** | Red Hat username with a valid license or free trial to Red Hat<br /> OpenShift Container Platform (RHOCP), which comes with<br /> necessary licenses for Red Hat Enterprise Linux (RHEL) and<br /> Red Hat CoreOS (RHCOS). | redhat.user
 **env.redhat.password** | Password to Red Hat above user's account. Used to auto-attach<br /> necessary subscriptions to KVM Host, bastion VM, and pull live<br /> images for OpenShift. | rEdHatPa$s!
+**env.redhat.manage_subscription** | In case that the supscription of the OS should be attached. Possible values: no/attach | attach
```
jacobemery, line 51:
```suggestion
**env.redhat.attach_subscription** | True or False. Would you like to subscribe the server with Red Hat? | True 
```

inventories/default/group_vars/all.yaml.template
```diff
@@ -44,6 +44,7 @@ env:
   redhat:
     username: #X
     password: #X
+    manage_subscription: attach
```
jacobemery, line 47:
```suggestion
    attach_subscription: true
```

inventories/default/group_vars/all.yaml.template
```diff
@@ -11,7 +11,7 @@ env:
 # Section 2 - LPAR(s)
   z:
     high_availability: False
-    ip_forward: #X
+    ip_forward: 0
```
jacobemery, line 14:
```suggestion
    ip_forward: "{{ 1 if network_mode | upper == 'NAT' else 0 }}"
```
What do you think about changing to this?

jacobemery: One less variable to forget to change when using NAT.

playbooks/3_setup_kvm_host.yaml
```diff
@@ -71,7 +71,7 @@
   vars:
     packages: "{{ env.pkgs.kvm }}"
   roles:
-    - { role: attach_subscription, when: env.redhat.username is defined and env.redhat.password is defined }
+    - { role: attach_subscription, when:  env.redhat.manage_subscription | upper == 'ATTACH' }
```
jacobemery, line 74:
```suggestion
    - { role: attach_subscription, when:  env.redhat.attach_subscription }
```

jacobemery (changes requested):
Suggestions to go along with previous comments about changing manage_subscription to a boolean value. Let me know what you think!

playbooks/3_setup_kvm_host.yaml
AmadeusPodvratnik: Done.

inventories/default/group_vars/all.yaml.template
AmadeusPodvratnik: This make sense ...

playbooks/3_setup_kvm_host.yaml
```diff
@@ -71,7 +71,7 @@
   vars:
     packages: "{{ env.pkgs.kvm }}"
   roles:
-    - { role: attach_subscription, when: env.redhat.username is defined and env.redhat.password is defined }
+    - { role: attach_subscription, when:  env.redhat.attach_subscription }
```
jacobemery, line 74: I just realized this role is also used in 5_setup_bastion, so the conditional would have to be changed there as well. 

But, how about instead of putting the conditional at the role-level as we have here and in playbook 5, we put it at the task-level (roles/attach_subscription/tasks/main.yaml) so that way we can remove the conditional from the playbook-level and we don't have the conditional in multiple places. Any future changes would need to be made in only one place, and it makes the playbooks cleaner as well.

jacobemery (changes requested):
One more comment for you. Suggesting a change to the placement of the conditional for attach_subscription. 

I did see the DCO checks failed. If you could check that out as well.

Otherwise, looks good! Once those are done we can merge.

AmadeusPodvratnik:
At the end I was able to sign the commits (accepting changes via web leads commits without signoff).

jacobemery (approved):
Looks good!

PR #200, opened 2023-09-14
fix: pull updated gh-pages

jacobemery:
Hotfix to #196 that automates GitHub Pages deployments. Looks like the workflow failed in error because it didn't have the most recent version of the gh-pages branch. This PR will hopefully fix that issue.


AmadeusPodvratnik (approved):
Looks good..

PR #203, opened 2023-09-19
feat: Add support to create aarch64 (arm64) compute node

smolin-de:
Add support to create a aarch64 (arm64) compute node using create_compute_node.yaml playbook.


AmadeusPodvratnik (commented):
@smolin-de: You added a new function / file "print_ocp_co_status.yaml" but I do not see that you are addressing it (like the include_tasks for print_ocp_node_status). Is this part of an upcoming PR?

smolin-de:
> @smolin-de: You added a new function / file "print_ocp_co_status.yaml" but I do not see that you are addressing it (like the include_tasks for print_ocp_node_status). Is this part of an upcoming PR?

Good catch. I will remove this function.

smolin-de:
I removed the file "print_ocp_co_status.yaml"

PR #209, opened 2023-10-06
Add ipv6 support

AmadeusPodvratnik:
These changes are fixing ipv6 support. Introducing new variables for ipv6. ipv6 will be set default (use_ipv6 var is set to True).
Fixing attach_subscription issue.


AmadeusPodvratnik:
@jacobemery pls have a look. I make ipv6 adr. default but not sure. I think it makes sense to have ipv6 addresses available. Doc will updated accordingly.

AmadeusPodvratnik:
@smolin-de Can you pls have a quick look on this PR?

playbooks/3_setup_kvm_host.yaml
```diff
@@ -71,7 +71,7 @@
   vars:
     packages: "{{ env.pkgs.kvm }}"
   roles:
-    - { role: attach_subscription, when:  env.redhat.attach_subscription }
+    - { role: attach_subscription, when:  env.redhat.manage_subscription }
```
smolin-de, line 74: In line 47 you are still using the variable "env.redhat.attach_subscription" !

playbooks/5_setup_bastion.yaml
```diff
@@ -71,7 +71,7 @@
         tasks_from: initial-resolv.yaml
       when: env.bastion.options.dns is defined and env.bastion.options.dns
   roles:
-    - { role: attach_subscription, when: env.redhat.username is defined and env.redhat.password is defined }
+    - { role: attach_subscription, when: env.redhat.manage_subscription }
```
smolin-de, line 74: In line 47 you are still using "env.redhat.attach_subscription"

docs/set-variables-group-vars.md
```diff
@@ -81,10 +84,13 @@
...
+**env.cluster.networking.base_domain** | The site name, where is the cluster being hosted? This will be combined with the metadata_name<br /> and hostnames to create FQDNs.  | host.com
+**env.bastion.networking.ipv6_gateway** | IPv6 of he bastion's gateway server. | fd00::1
+**env.bastion.networking.ipv6_prefix** | IPv6 prefix. | 64
 **env.cluster.networking.nameserver1** | IPv4 address that the cluster get its hostname resolution from. If env.bastion.options.dns<br /> is True, this should be the IP address of the bastion. | 192.168.10.200
 **env.cluster.networking.nameserver2** | <b>(Optional)</b> A second IPv4 address will the cluster get its hostname resolution from? If env.bastion.options.dns<br /> is True, this should be left commented out. | 192.168.10.201
 **env.cluster.networking.forwarder** | What IPv4 address will be used to make external DNS calls for the cluster? Can use 1.1.1.1 or 8.8.8.8 as defaults. | 8.8.8.8
+**env.cluster.networking.interface** | Name of the networking interface on the bastion from Linux's perspective. Most likely enc1. | enc1
```
smolin-de, line 93: As I know "enc1" is used on s390x VMs. On arm64 and x86_64 VM's you have a different interface name suffix.
Right now the value in this variable is used on all architectures, which will break Multi-Arch setup with KVM.

smolin-de (changes requested):

playbooks/5_setup_bastion.yaml
AmadeusPodvratnik: Fixed

docs/set-variables-group-vars.md
AmadeusPodvratnik: Thx for the comment. I did the same like for bastion (there is also enc1 used as a interface name). Is it there the same problem? Within the all.yaml you need to specify the interface and the user can put the correct one. Is this not possible on x86?

smolin-de: The issue is if you are using multiple architectures and multiple systems. e.g. on my x86 server the VM interface name is : enp1s0. On a different x86 you may have to use use a different interface name. 
Right now we are not specifying the interface name. It works, because we have only one interface which is used.
I recommend to remove the "**env.cluster.networking.interface** " variable, and focus only required ipv6 changes.

AmadeusPodvratnik: Well ... the problem is that if I do not specify the interfaces using ipv5 then the network will not come up properly. I tried to use no ipv4 interface and ipv6 interface and ended up in network issues where VM is not accessible or does not get a ipv6 adr. Therefor I specified the interface.
Is this a day 1 issue, a day 2 issue or both?
Is it necessary to specify different interface names in case that the cluster is spread across multiple LPARs?

smolin-de: Day 1 is currently used only on one architecture. So, the issue which I have described above exists when you use Day1 and Day 2 operations in one pipeline.
A potential solution would be to define the VM network interface for each LPAR//server.

smolin-de (approved):

docs/set-variables-group-vars.md
AmadeusPodvratnik: @smolin-de As discussed ... I add the interface to the day2 operation.

Issue #182, opened 2023-07-29
Task 'Check internal cluster DNS resolution'  fails, when external DNS is used


smolin-de:
When I use an external DNS server, by setting "env.bastion.options.dns: false" then the following tasks from playbooks/5_setup_bastion.yaml will fail:

```
TASK [check_dns : Check internal cluster DNS resolution for the bastion] *******************************************************************************
fatal: [bastion]: FAILED! => {"changed": true, "cmd": "dig +short bastion.lnxero1.boe | tail -n1", "delta": "0:00:00.010796", "end": "2023-07-29 07:12:14.980317", "failed_when_result": true, "msg": "", "rc": 0, "start": "2023-07-29 07:12:14.969521", "stderr": "", "stderr_lines": [], "stdout": "", "stdout_lines": []}
```

```
TASK [check_dns : Check internal cluster DNS resolution for bootstrap] *********************************************************************************
fatal: [bastion]: FAILED! => {"changed": true, "cmd": "dig +short bootstrap.ocp3.lnxero1.boe | tail -n1", "delta": "0:00:00.011422", "end": "2023-07-29 07:12:23.012899", "failed_when_result": true, "msg": "", "rc": 0, "start": "2023-07-29 07:12:23.001477", "stderr": "", "stderr_lines": [], "stdout": "172.23.232.224", "stdout_lines": ["172.23.232.224"]}
```

Issue #166, opened 2023-06-04
Playbook 5 says dig validation of DNS fails, but dig lookup commands are succesful


ryoung1us:
Ansible Output:

TASK [dns : Template out bastion's resolv.conf file, replacing initial resolv.conf] ***************************************************************************************
skipping: [bastion.ocp1.ibm.com]

TASK [dns : Restart network to update changes made to /etc/resolv.conf] ***************************************************************************************************
skipping: [bastion.ocp1.ibm.com]

TASK [check_dns : Check internal cluster DNS resolution for the bastion] **************************************************************************************************
changed: [bastion.ocp1.ibm.com]

TASK [check_dns : Check internal cluster DNS resolution for external API and apps services] *******************************************************************************
failed: [bastion.ocp1.ibm.com] (item=api.ocp1.ibm.com) => {"ansible_loop_var": "item", "changed": true, "cmd": "dig +short api.ocp1.ibm.com | tail -n1", "delta": "0:00:00.007357", "end": "2023-06-03 20:20:19.514264", "failed_when_result": true, "item": "api.ocp1.ibm.com", "msg": "", "rc": 0, "start": "2023-06-03 20:20:19.506907", "stderr": "", "stderr_lines": [], "stdout": "9.76.61.80", "stdout_lines": ["9.76.61.80"]}
failed: [bastion.ocp1.ibm.com] (item=apps.ocp1.ibm.com) => {"ansible_loop_var": "item", "changed": true, "cmd": "dig +short apps.ocp1.ibm.com | tail -n1", "delta": "0:00:00.006574", "end": "2023-06-03 20:20:19.777103", "failed_when_result": true, "item": "apps.ocp1.ibm.com", "msg": "", "rc": 0, "start": "2023-06-03 20:20:19.770529", "stderr": "", "stderr_lines": [], "stdout": "9.76.61.80", "stdout_lines": ["9.76.61.80"]}
[...]

PR #212, opened 2023-10-12
fix: Temporary ignore some DNS checks

smolin-de:
There are several open issues related to DNS. DNS setup needs to be verified. We ignore some DNS check results for now, because of issue 166, 182.

### DIFF
--- a/playbooks/5_setup_bastion.yaml
+++ b/playbooks/5_setup_bastion.yaml
@@ -76,7 +76,7 @@
     - ssh_ocp_key_gen
     - set_firewall
     - { role: dns, when: env.bastion.options.dns }
-    - check_dns
+    - { role: check_dns, when env.bastion.options.dns is defined and env.bastion.options.dns }
     - { role: haproxy, when: env.bastion.options.loadbalancer.on_bastion }
     - httpd
 


AmadeusPodvratnik:
@smolin-de I looked into #182 and it looks to me that using 
- { role: check_dns, when env.bastion.options.dns is defined and env.bastion.options.dns} instead of 
- check_dns
in 5_setup_bastion playbook will solve the issue of an external dns, correct?
But looking in #166 we can fix it differently...
As far as I understood the issues is, that the output of services_lookup.stdout is different from bastion ip (check in check-dns role: (env.bastion.options.loadbalancer.on_bastion == True) and (env.bastion.networking.ip != services_lookup.stdout)). 
If we would like to have an api dns check we need to introduce the IP of the dns server (in all.yaml).

AmadeusPodvratnik (changes requested):
Pls change the role call in 5_setup_bastion playbook:
{ role: check_dns, when env.bastion.options.dns is defined and env.bastion.options.dns} instead of
check_dns

AmadeusPodvratnik (commented):
Looks good, thank you!

AmadeusPodvratnik (approved):

PR #214, opened 2023-10-13
Fix: Added os flag for virt-install command and changed hypershift binary name 

veera-damisetti:
- Added os flag for virt-install command as it is a required flag in rhel9 
- Changed hypershift binary name to hcp
- added new flags in cluster creation command


AmadeusPodvratnik (changes requested):
Hi @veera-damisetti
Not sure why you commented the ip_forward out because it has a dependency on NAT.
Can you pls explain why?
You removed libvirt-devel from package install. Is it not required anymore (in general)?
You introduced the --osinfo flag and the question is, is this flag available on rhcos 8?

veera-damisetti:
@AmadeusPodvratnik 

I did 2 changes in the code which is outside hypershift ( Debug purpose ) 

1. commented the ip_forward

It has one variable network_mode , which is breaking my flow with error 
network_mode is undefined 
just to bypass that issue to break my flow temporarily ---- but need to report this issue.

2. libvirt-devel 

this package is not available in rhel9 
so this is also breaking for rhel9

these 2 points actually i thought of discussing in our coordination call .
just for debug purpose i added in this  PR

veera-damisetti:
@AmadeusPodvratnik
Removed rhel9 related changes from this PR as they need to be discussed separately
Keeping hypershift related changed and removed DO-NOT-MERGE in the name 
please review them
thanks ..!

AmadeusPodvratnik (approved):
Looks good so far ...

Issue #223, opened 2023-11-16
haproxy config file does not contain day-2 compute nodes


smolin-de:
Additional compute nodes which are added with the the "playbooks/create_compute_node.yaml" with a day-2 install,
are not added to the "haproxy.cfg" file.

```
cat  /etc/haproxy/haproxy.cfg
...
listen ingress-router-443
  bind *:443
  mode tcp
  balance source
  #443 section
  server worker-2 worker-2.ocp1.test.multiarch:443 check inter 1s
  server worker-1 worker-1.ocp1.test.multiarch:443 check inter 1s
listen ingress-router-80
  bind *:80
  mode tcp
  balance source
  #80 section
  server worker-1 worker-1.ocp1.test.multiarch:80 check inter 1s
  server worker-2 worker-2.ocp1.test.multiarch:80 check inter 1s
  
```

smolin-de:
This defect can be closed, because of above merged PR.

AmadeusPodvratnik:
As @smolin-de mentioned above. Close defect.

PR #224, opened 2023-11-22
fix: Update HAProxy.cfg during day-2 compute node install

smolin-de:
This patch updates the HAProxy.cfg file during compute node install and removal.

### DIFF
--- a/playbooks/create_compute_node.yaml
+++ b/playbooks/create_compute_node.yaml
@@ -99,3 +99,8 @@
     - role: approve_certs
       approve_certs_cleanup: true
     - role: print_node_status
+
+    - role: haproxy_update
+      when: env.bastion.options.loadbalancer.on_bastion
+      param_haproxy_cmd: add
+      param_haproxy_hostname: "{{ day2_compute_node.vm_hostname }}"
--- a/playbooks/delete_compute_node.yaml
+++ b/playbooks/delete_compute_node.yaml
@@ -35,6 +35,11 @@
       # Some tasks will be delegated to the bastion host and target KVM host
       param_compute_node: "{{ day2_compute_node }}"
 
+    - role: haproxy_update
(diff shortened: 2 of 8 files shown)


smolin-de:
This patch resolves issue #223

AmadeusPodvratnik (commented):
Hi @smolin-de, thx for the PR. I have following questions:
- Why you are opening 1936 for tcp on the firewall?
- You add haproxy role to reinstall cluster and for day2 operations but in the general flow the haproxy is configured during 5_setup_bastion playbook. So for reinstall cluster nothing should change or am I wrong.

smolin-de:
> Hi @smolin-de, thx for the PR. I have following questions:
> 
>     * Why you are opening 1936 for tcp on the firewall?
> 
>     * You add haproxy role to reinstall cluster and for day2 operations but in the general flow the haproxy is configured during 5_setup_bastion playbook. So for reinstall cluster nothing should change or am I wrong.


Q -  Why you are opening 1936 for tcp on the firewall?
A - The current HAProxy config file uses port 1936 to access HAProxy statistics, but this port is not enabled in the firewall.
      I enabled the port to access HAProxy statistics.  But, it is not required for the fix. We could or should create a separate patch for the port issue. I will remove the fort change from this patch.
 
 Q - You add haproxy role to reinstall cluster and for day2 operations but in the general flow the haproxy is configured during 5_setup_bastion playbook. So for reinstall cluster nothing should change or am I wrong.
A - If you install a cluster and add day-2 compute nodes, then your HAproxy config will contain all compute nodes. If you now reinstall the cluster and want to have a different day-2 setup, then you run into an issue. Because, the HAProxy config file contains the old setup with the old day-2 compute nodes. Therefore, I am resetting the HAProxy config file to the default setup without day-2 compute node entries.

AmadeusPodvratnik (approved):
Thank you ... looks good.

PR #226, opened 2023-11-25
Added changes to save HCP password,skipped pakages for rhel9 for HCP 

veera-damisetti:
- Updated to save the HCP password on bastion - kubadmin-password
- Fix  for creating bastion for HCP 
- Added change in hcp prerequisite installation , to skip devel packages if rhlel version >= 9 
- added change to delete the expect file after usage


AmadeusPodvratnik (commented):
Looks good so far. One question (or two). If you save the password separately do you need to document where the password will be and will this password used in a later step?

veera-damisetti:
yeah i am saving the password where kubeconfig is there 
which is the default thing i guess . 
we still need to mention the location of where the password is .?

AmadeusPodvratnik:
> yeah i am saving the password where kubeconfig is there which is the default thing i guess . we still need to mention the location of where the password is .?

Yes please. Just in case that something went wrong the user should know where security relevant data is located.

AmadeusPodvratnik (approved):
Looks good so far ...

PR #234, opened 2024-01-03
fix: Execute function when required variable is defined

smolin-de:
Execute function when required variable is defined.


AmadeusPodvratnik (approved):
Hi @smolin-de , thx for the PR ... looks good so far.

PR #236, opened 2024-01-16
Adding post mirror actions for oc-mirror.

mohammedzee1000:
This allows user to perform actions on the mirror host after
oc mirror command is run.

The replace action allows user to change the mapping.txt and repush
images which is useful is in cases where multiple endpoints exist
for registry and some of them are hidden such as when user wants to
mirror pre-release operators using iib index images.

Part of #143 

Signed-off-by: Mohammed Zeeshan Ahmed <mohammed.zee1000@gmail.com>


AmadeusPodvratnik (approved):
Looks good so far ... thank you.

PR #237, opened 2024-01-17
fix: Delete node when node name contains domain name as suffix

smolin-de:
Delete node does not work when ocp is using full guqlifed node names (with domain name suffix).


AmadeusPodvratnik (approved):
Looks good so far.

PR #239, opened 2024-02-15
feat: Add support for ARM KVM server

smolin-de:
This patch contains several changes which are required to install
OCP on a ARM server.
The new variable 'vcpu_model_option' can be used to configure the
CPU model and CPU features exposed to the guest. The default CPU
model 'host' is currently not supported on ARM servers and must be
defined to the desired value.

### DIFF
--- a/roles/create_bastion/templates/bastion-ks.cfg.j2
+++ b/roles/create_bastion/templates/bastion-ks.cfg.j2
@@ -54,11 +54,19 @@ bootloader --append="crashkernel=auto" --location=mbr --boot-drive=vda
 clearpart --all --initlabel --drives=vda
 
 # Disk partitioning information
+{% if env.install_config.control.architecture == 'arm64' %}
+# TODO: Special setup for arm required, becase our arm server requires /boot/efi partition with efi file system
+ignoredisk --only-use=vda
+# System bootloader configuration
+bootloader --location=mbr --boot-drive=vda
+autopart
+{% else %}
 part /boot --fstype="xfs" --asprimary --ondisk=vda --size=1024
 part pv.01 --fstype="lvmpv" --grow --size=1 --ondisk=vda
 volgroup vgsystem --pesize=4096 pv.01
 logvol swap --fstype=swap --name=swap --vgname=vgsystem --size={{ env.bastion.resources.swap }}
 logvol / --fstype=xfs --name=root --vgname=vgsystem --size=1 --grow
+{% endif %}
 
 # Packages selection
 %packages --multilib --ignoremissing
--- a/roles/create_bastion/templates/bastion-ks.cfg.j2
+++ b/roles/create_bastion/templates/bastion-ks.cfg.j2
@@ -55,7 +55,7 @@ clearpart --all --initlabel --drives=vda
 
 # Disk partitioning information
 {% if env.install_config.control.architecture == 'arm64' %}
-# TODO: Special setup for arm required, becase our arm server requires /boot/efi partition with efi file system
+# TODO: Special setup for arm required, because our arm server requires /boot/efi partition with efi file system
 ignoredisk --only-use=vda
 # System bootloader configuration
 bootloader --location=mbr --boot-drive=vda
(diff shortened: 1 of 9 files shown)


roles/create_compute_nodes/tasks/main.yaml
```diff
@@ -4,14 +4,18 @@
...
     virsh destroy {{ env.cluster.nodes.compute.vm_name[i] }} || true
-    virsh undefine {{ env.cluster.nodes.compute.vm_name[i] }} --remove-all-storage || true
+    virsh undefine {{ env.cluster.nodes.compute.vm_name[i] }} --remove-all-storage --nvram || true
     virt-install \
     --name {{ env.cluster.nodes.compute.vm_name[i] }} \
     --osinfo detect=on,name={{ ('rhel8.6') if rhcos_os_variant is not defined else (rhcos_os_variant) }} \
     --autostart \
     --disk pool={{ env.cluster.networking.metadata_name }}-vdisk,size={{ env.cluster.nodes.compute.disk_size }}  \
     --ram {{ env.cluster.nodes.compute.ram }} \
+    {% if env.cluster.nodes.compute.vcpu_model_option is defined %}
```
AmadeusPodvratnik, line 14: You wrote in the description:
(Optional) Configure the CPU model and CPU features exposed to the guest | --cpu host
Make it sense to make --cpu host as default and change the all.yaml.template? Then you don't need the if statement.

roles/create_bastion/templates/bastion-ks.cfg.j2
```diff
@@ -54,11 +54,19 @@ bootloader --append="crashkernel=auto" --location=mbr --boot-drive=vda
 clearpart --all --initlabel --drives=vda
 
 # Disk partitioning information
+{% if env.install_config.control.architecture == 'arm64' %}
```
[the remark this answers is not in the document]
smolin-de: why we should add a additional parameter ?

roles/create_compute_nodes/tasks/main.yaml
smolin-de: The default option is already "--cpu host"
If the user does not specify the variable "--cpu host " is used

e.g.
    {% if env.cluster.nodes.bootstrap.vcpu_model_option is defined %}
    {{ env.cluster.nodes.bootstrap.vcpu_model_option }} \
    {% else %}
    --cpu host \
    {% endif %}

smolin-de: Keep in mind that "--cpu host" is not supported on our ARM server.
Thats why I have added this variable.

smolin-de: For arm server you have to set the variable to "" or define a value which is supported.

roles/create_bastion/templates/bastion-ks.cfg.j2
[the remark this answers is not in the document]
AmadeusPodvratnik: I don't say that you should add another parameter but add the arm64 option to the description of the env.install_config.control.architecture parameter.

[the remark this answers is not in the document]
smolin-de: I add update. Please have a look.

roles/create_compute_nodes/tasks/main.yaml
smolin-de: Yes, addding --cpu host" to aall.yaml.template is an option.  and we would get rid of the if statement.
But, in this case existing user configuration must be adapted to the new template file.
Both option are possible.

I can change the code, and remove the if statement if you like ?

AmadeusPodvratnik: @smolin-de Yes, pls change the code. We discussed yesterday how we would like to proceed here. We will have a What's (Readme) pointing to the additional variables.

smolin-de: Code changes done.

AmadeusPodvratnik (approved):
@smolin-de thx... looks good so far.

PR #240, opened 2024-02-19
feat: Add DHCP support

smolin-de:
With this patch you can boot bastion, bootstrap node, control node and compute node via DHCP service.
Infra nodes are not supported yet.


AmadeusPodvratnik (approved):
Looks good so far. Thank you.

PR #241, opened 2024-02-28
Lpar integration - installation on LPAR as nodes 

sandisamp:
Create integration of LPAR into playbook - 6_create_nodes

Accordingly tasks and roles have been created.

**Note: This PR is a work in progress. Created for initial reviews and not to make the review process too long. Will update detailed description when PR is ready.**

### DIFF
--- a/docs/set-variables-host-vars.md
+++ b/docs/set-variables-host-vars.md
@@ -15,8 +15,11 @@
 :--- | :--- | :---
 **networking.hostname** | The hostname of the LPAR with RHEL installed natively (the KVM host). | kvm-host-01
 **networking.ip** | The IPv4 address of the LPAR with RHEL installed natively (the KVM host). | 192.168.10.2
+**networking.ipv6** | IPv6 address for the bastion if use_ipv6 variable is 'True'. | fd00::3
 **networking.subnetmask** | The subnet that the LPAR resides in within your network. | 255.255.255.0
 **networking.gateway** | The IPv4 address of the gateway to the network where the KVM host resides. | 192.168.10.0
+**networking.ipv6_gateway** | IPv6 of he bastion's gateway server. | fd00::1
+**networking.ipv6_prefix** | IPv6 prefix. | 64
 **networking.nameserver1** | The IPv4 address from which the KVM host gets its hostname resolved. | 192.168.10.200
 **networking.nameserver2** | <b>(Optional)</b> A second IPv4 address from which the KVM host can get its hostname<br /> resolved. Used for high availability. | 192.168.10.201
 **networking.device1** | The network interface card from Linux's perspective. Usually enc and then a number that comes<br /> from the dev_num of the network adapter. | enc100
@@ -25,8 +28,11 @@
 
 ## Important Note
 * You can skip the rest of the variables on this page IF you are using existing LPAR(s) that has RHEL already installed.
+* If you are installing an LPAR based cluster then the information below must be provided and are not optional. You must create a host file corresponding to each lpar node.
     * Since this is how most production deployments on-prem are done on IBM zSystems, these variables have been marked as optional. 
     * With pre-existing LPARs with RHEL installed, you can also skip [1_create_lpar.yaml](https://github.com/IBM/Ansible-OpenShift-Provisioning/blob/main/playbooks/1_create_lpar.yaml) and [2_create_kvm_host.yaml](https://github.com/IBM/Ansible-OpenShift-Provisioning/blob/main/playbooks/2_create_kvm_host.yaml) playbooks. Make sure to still do [0_setup.yaml](https://github.com/IBM/Ansible-OpenShift-Provisioning/blob/main/playbooks/0_setup.yaml) first though, then skip to [3_setup_kvm_host.yaml](https://github.com/IBM/Ansible-OpenShift-Provisioning/blob/main/playbooks/3_setup_kvm_host.yaml)
+    * In the scenario of lpar based installation you can skip [1_create_lpar.yaml](https://github.com/IBM/Ansible-OpenShift-Provisioning/blob/main/playbooks/1_create_lpar.yaml) and [2_create_kvm_host.yaml](https://github.com/IBM/Ansible-OpenShift-Provisioning/blob/main/playbooks/2_create_kvm_host.yaml). You can also optionally skip [3_setup_kvm_host.yaml](https://github.com/IBM/Ansible-OpenShift-Provisioning/blob/main/playbooks/3_setup_kvm_host.yaml) and [4_create_bastion.yaml](https://github.com/IBM/Ansible-OpenShift-Provisioning/blob/main/playbooks/3_setup_kvm_host.yaml) unless you are planning on having the bastion on the same host.
+    * In case of lpar based installation one is expected to have a tessia live disk accessible by the lpar nodes for network boot. The details of which are to be filled in section #7 below. The steps to create a tessia livedisk can be found [here](https://gitlab.com/tessia-project/tessia-baselib/-/blob/master/doc/users/live_image.md).
 
 ## 2 - (Optional) CPC & HMC
 **Variable Name** | **Description** | **Example**
@@ -82,3 +88,10 @@
 **lpar.storage_group_2_.storage_wwpn** | <b>(Optional)</b> World-wide port numbers for storage group. Use provided list formatting. | 500708680235c3f0<br />500708680235c3f1<br />500708680235c3f2<br />500708680235c3f3
 **lpar.storage_group_2_.dev_num** | <b>(Optional)</b> The logical device number of the Host Bus Adapter (HBA) for the storage group. | C001
 **lpar.storage_group_2_.lun_name** | <b>(Optional)</b> he Logical Unit Numbers (LUN) that points to a specific virtual disk<br /> behind the WWPN. | 4200569309ahhd240000000000000c001
+
+## 7 - (Optional) Livedisk info
+**Variable Name** | **Description** | **Example**
+:--- | :--- | :---
+**lpar.livedisk.livedisktype** | <b>(Optional)</b> Storage type. DASD is the only tested type as of now. | dasd
+**lpar.livedisk.devicenr** | <b>(Optional)</b> the device no of the DASD live disk | c6h1
+**lpar.livedisk.livedisk_root_pass** | <b>(Optional)</b> root password for the livedisk | p@ssword
--- a/inventories/default/host_vars/KVMhostname1-here.yaml.template
+++ b/inventories/default/host_vars/KVMhostname1-here.yaml.template
@@ -2,8 +2,11 @@
 networking:
   hostname: #X
   ip: #X
+  ipv6: #X
   subnetmask: #X
+  ipv6_prefix: #X
   gateway: #X
+  ipv6_gateway: #X
   nameserver1: #X
 #  nameserver2: 
   device1: #X
@@ -14,9 +17,11 @@ storage:
 
 ##############################################################
 # Variables below this point only need to be filled out if   #
-# env.z.lpar1.create is True. Meaning, if the LPARs you will #
+# env.z.lpar1.create is True or if you are doing a lpar      #
+# based installation. Meaning, if the LPARs you will         #
 # be using as KVM host(s) already exist and have RHEL        #   
-# installed, the variables below are not required.           #
+# installed, the variables below are not required. However   #
+# for lpar based installation you must provide below vars    #
 ##############################################################
 
 # Section 2 - CPC & HMC
@@ -81,4 +86,10 @@ lpar:
 #      - 
 #      - 
 #    dev_num: 
-#    lun_name: 
+#    lun_name:
+
+# Section 7 -live disk info
+  livedisk:
+    livedisktype: #X 
+    devicenr: #X
+    livedisk_root_pass: #X 
--- a/inventories/default/host_vars/KVMhostname2-here.yaml.template
+++ b/inventories/default/host_vars/KVMhostname2-here.yaml.template
@@ -2,8 +2,11 @@
 networking:
   hostname: #X
   ip: #X
+  ipv6: #X
   subnetmask: #X
+  ipv6_prefix: #X
   gateway: #X
+  ipv6_gateway: #X
   nameserver1: #X
 #  nameserver2: 
   device1: #X
@@ -14,9 +17,11 @@ storage:
 
 ##############################################################
 # Variables below this point only need to be filled out if   #
-# env.z.lpar2.create is True. Meaning, if the LPARs you will #
+# env.z.lpar2.create is True or if you are doing a lpar      #
+# based installation. Meaning, if the LPARs you will         #
 # be using as KVM host(s) already exist and have RHEL        #   
-# installed, the variables below are not required.           #
+# installed, the variables below are not required. However   #
+# for lpar based installation you must provide below vars    #
 ##############################################################
 
 # Section 2 - CPC & HMC
@@ -82,3 +87,9 @@ lpar:
 #      - 
 #    dev_num: 
 #    lun_name: 
+
+# Section 7 -live disk info
+  livedisk:
+    livedisktype: #X 
+    devicenr: #X
+    livedisk_root_pass: #X
--- a/inventories/default/host_vars/KVMhostname3-here.yaml.template
+++ b/inventories/default/host_vars/KVMhostname3-here.yaml.template
@@ -2,8 +2,11 @@
 networking:
   hostname: #X
   ip: #X
+  ipv6: #X
   subnetmask: #X
+  ipv6_prefix: #X
   gateway: #X
+  ipv6_gateway: #X
   nameserver1: #X
 #  nameserver2: 
   device1: #X
@@ -14,9 +17,11 @@ storage:
 
 ##############################################################
 # Variables below this point only need to be filled out if   #
-# env.z.lpar3.create is True. Meaning, if the LPARs you will #
+# env.z.lpar3.create is True or if you are doing a lpar      #
+# based installation. Meaning, if the LPARs you will         #
 # be using as KVM host(s) already exist and have RHEL        #   
-# installed, the variables below are not required.           #
+# installed, the variables below are not required. However   #
+# for lpar based installation you must provide below vars    #
 ##############################################################
 
 # Section 2 - CPC & HMC
@@ -82,3 +87,9 @@ lpar:
 #      - 
 #    dev_num: 
 #    lun_name: 
+
+# Section 7 -live disk info
+  livedisk:
+    livedisktype: #X 
+    devicenr: #X
+    livedisk_root_pass: #X
--- a/playbooks/0_setup.yaml
+++ b/playbooks/0_setup.yaml
@@ -39,7 +39,7 @@
       tags: lpar_check
       stat:
         path: "{{ inventory_dir }}/host_vars/{{ env.z.lpar1.hostname }}.yaml"
-      when: env.z.lpar1.create == True
+      when: env.z.lpar1.hostname is defined
       register: lpar_host_vars
       failed_when: lpar_host_vars.stat.exists == False
 
--- a/playbooks/0_setup.yaml
+++ b/playbooks/0_setup.yaml
@@ -39,7 +39,7 @@
       tags: lpar_check
       stat:
         path: "{{ inventory_dir }}/host_vars/{{ env.z.lpar1.hostname }}.yaml"
-      when: env.z.lpar1.hostname is defined
+      when: env.z.lpar1.create == True
       register: lpar_host_vars
       failed_when: lpar_host_vars.stat.exists == False
 
--- a/playbooks/5_setup_bastion.yaml
+++ b/playbooks/5_setup_bastion.yaml
@@ -81,14 +81,7 @@
     - { role: haproxy, when: env.bastion.options.loadbalancer.on_bastion }
     - { role: sno_haproxy, when: env.bastion.options.loadbalancer.on_bastion and control_node_count | int == 1 }
     - httpd
-
-- name: Install tessia-baselib for lpar based installtions
-  hosts: bastion
-  tags: services, section_2
-  vars_files: 
-  - "{{ inventory_dir }}/group_vars/all.yaml"
-  roles:
-  - { role: install_tessia_baselib, when: env.installation_type|lower == "lpar" }
+    - { role: install_tessia_baselib, when: env.installation_type|lower == "lpar" }
 
 - hosts: bastion
   tags: services, section_2, openvpn
--- a/playbooks/5_setup_bastion.yaml
+++ b/playbooks/5_setup_bastion.yaml
@@ -82,6 +82,14 @@
     - { role: sno_haproxy, when: env.bastion.options.loadbalancer.on_bastion and control_node_count | int == 1 }
     - httpd
 
+- name: Install tessia-baselib for lpar based installtions
+  hosts: bastion
+  tags: services, section_2
+  vars_files: 
+  - "{{ inventory_dir }}/group_vars/all.yaml"
+  roles:
+  - { role: install_tessia_baselib, when: env.bastion.installation_type|lower|lower == "lpar" }
+
 - hosts: bastion
   tags: services, section_2, openvpn
   become: true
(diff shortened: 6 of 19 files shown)


AmadeusPodvratnik (changes requested):
Hi @sandisamp , thx for the PR. Following additonal comments:
- There is a description missing what support you are adding (lpar).
- You need to document what your are supporting (Classic vs. DPM, fcp device, etc). 
- You have additional variables defined which needed to be explained and described.
- In your boot_LPAR task I do not see a interface name and you are supporting ipv4 only ... is this correct?

playbooks/5_setup_bastion.yaml
```diff
@@ -140,7 +140,7 @@
       loop:
         - issued
         - private
-      when: env.z.high_availability == True
+      when: env.z.high_availability == True and env.bastion.type != "zvm" and env.bastion.type != "lpar"
```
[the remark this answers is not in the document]
smolin-de: @sandisamp 
if available, please add your code to install tessia and all other required packages.

roles/install_packages/tasks/main.yaml
```diff
@@ -14,6 +14,11 @@
   loop: "{{ packages }}"
   when: ansible_os_family != 'Darwin'
 
+- name:  Install tessia baselib
+  import_role:
+    name: install_tessia_baselib
+  when: env.bastion.installation_type == "lpar"
+
```
smolin-de, line 21: This will install tessia_baselib on my  Laptop, on all KVM hosts and on bastion Bastion host.
Why ?

playbooks/5_setup_bastion.yaml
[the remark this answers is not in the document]
sandisamp: The role to install Tessia is already written  for hyper shift use case. Added it here as well.

playbooks/5_setup_bastion.yaml
```diff
@@ -87,8 +87,8 @@
     openvpn_role: "server"
   roles:
     #- { role: robertdebock.bootstrap, tags: openvpn, when: env.z.high_availability == True }
-    - { role: robertdebock.epel, tags: openvpn, when: env.setup_openvpn == True and env.z.high_availability == True }
-    - { role: robertdebock.openvpn, tags: openvpn, when: env.setup_openvpn == True and env.z.high_availability == True }
```
AmadeusPodvratnik, line 91: The check for env.setup_openvpn should not be removed. Pls add this check.

playbooks/5_setup_bastion.yaml
```diff
@@ -99,7 +99,7 @@
       file:
         state: directory
         path: tmp
-      when: env.setup_openvpn == True and env.z.high_availability == True
+      when: env.z.high_availability == True and env.bastion.installation_type != "zvm" and env.bastion.installation_type != "lpar"
```
AmadeusPodvratnik, line 102: The check for env.setup_openvpn should not be removed. Pls add this check.

playbooks/5_setup_bastion.yaml
```diff
@@ -116,7 +116,7 @@
         - issued/client.crt
         - private/client.key
         - ta.key
-      when: env.setup_openvpn == True and env.z.high_availability == True
+      when: env.z.high_availability == True and env.bastion.installation_type != "zvm" and env.bastion.installation_type != "lpar"
```
AmadeusPodvratnik, line 119: The check for env.setup_openvpn should not be removed. Pls add this check.

playbooks/5_setup_bastion.yaml
```diff
@@ -129,7 +129,7 @@
   pre_tasks:
     - name: Gather facts.
       setup:
-      when: env.setup_openvpn == True and env.z.high_availability == True
+      when: env.z.high_availability == True and env.bastion.installation_type != "zvm" and env.bastion.installation_type != "lpar"
```
AmadeusPodvratnik, line 132: The check for env.setup_openvpn should not be removed. Pls add this check.

playbooks/5_setup_bastion.yaml
```diff
@@ -140,7 +140,7 @@
       loop:
         - issued
         - private
-      when: env.setup_openvpn == True and env.z.high_availability == True
+      when: env.z.high_availability == True and env.bastion.installation_type != "zvm" and env.bastion.installation_type != "lpar"
```
AmadeusPodvratnik, line 143: The check for env.setup_openvpn should not be removed. Pls add this check.

playbooks/5_setup_bastion.yaml
```diff
@@ -153,10 +153,11 @@
         - client.crt
         - client.key
         - ta.key
-      when: env.setup_openvpn == True and env.z.high_availability == True
+
+      when: env.z.high_availability == True and env.bastion.installation_type != "zvm" and env.bastion.installation_type != "lpar"
   roles:
-    - { role: robertdebock.epel, tags: openvpn, when: env.setup_openvpn == True and env.z.high_availability == True }
-    - { role: robertdebock.openvpn, tags: openvpn, when: env.setup_openvpn == True and env.z.high_availability == True }
+    - { role: robertdebock.epel, tags: openvpn, when: env.z.high_availability == True and env.bastion.installation_type != "zvm" and env.bastion.installation_type != "lpar"  }
```
AmadeusPodvratnik, line 159: The check for env.setup_openvpn should not be removed. Pls add this check.

playbooks/5_setup_bastion.yaml
```diff
@@ -167,7 +168,7 @@
       file:
         state: absent
         path: tmp
-      when: env.setup_openvpn == True and env.z.high_availability == True
```
AmadeusPodvratnik, line 170: The check for env.setup_openvpn should not be removed. Pls add this check.

roles/boot_LPAR/tasks/main.yaml
```diff
@@ -0,0 +1,70 @@
...
+          --netset_ip {{ node.networking.ip }} \ 
+          --netset_gateway {{ node.networking.gateway }} \
+          --netset_network_type osa \
+          --netset_network_device "{{ node.networking.device1 }},{{ node.networking.device2 }},{{ node.networking.device3 }}" \
+          --netset_password {{ node.lpar.livedisk.livedisk_root_pass }} \
+          --netset_dns "{{ node.networking.nameserver1 }},{{ node.networking.nameserver2 }}" \
+          --log_level DEBUG \
+          --cmdline 'rd.neednet=1 console=ttysclp0 coreos.inst.install_dev=sda coreos.live.rootfs_url=http://{{ env.bastion.networking.ip }}:8080/bin/{{ rhcos_live_rootfs }} coreos.inst.ignition_url=http://{{ env.bastion.networking.ip }}:8080/ignition/{{ ignition }}.ign ip={{ node.networking.ip }}::{{ node.networking.gateway }}:255.255.0.0:{{ node.networking.hostname }}.{{ env.cluster.networking.metadata_name }}.{{ env.cluster.networking.base_domain }}::none nameserver={{ env.bastion.networking.ip }} cio_ignore=all,!condev zfcp.allow_lun_scan=0 rd.znet=qeth,{{ node.networking.device1 }},{{ node.networking.device2 }},{{ node.networking.device3 }},layer2=1 {{ rd_zfcp_string }}'
```
AmadeusPodvratnik, line 47: There is no support for ipv6 (which is available for KVM) is this as designed? If yes, when do you plan to add ipv6 support?

roles/boot_LPAR/tasks/main.yaml
```diff
@@ -0,0 +1,70 @@
...
+          --netset_gateway {{ node.networking.gateway }} \
+          --netset_network_type osa \
+          --netset_network_device "{{ node.networking.device1 }},{{ node.networking.device2 }},{{ node.networking.device3 }}" \
+          --netset_password {{ node.lpar.livedisk.livedisk_root_pass }} \
+          --netset_dns "{{ node.networking.nameserver1 }},{{ node.networking.nameserver2 }}" \
+          --log_level DEBUG \
+          --cmdline 'rd.neednet=1 console=ttysclp0 coreos.inst.install_dev=sda coreos.live.rootfs_url=http://{{ env.bastion.networking.ip }}:8080/bin/{{ rhcos_live_rootfs }} coreos.inst.ignition_url=http://{{ env.bastion.networking.ip }}:8080/ignition/{{ ignition }}.ign ip={{ node.networking.ip }}::{{ node.networking.gateway }}:255.255.0.0:{{ node.networking.hostname }}.{{ env.cluster.networking.metadata_name }}.{{ env.cluster.networking.base_domain }}:{{ env.cluster.networking.interface }}:none nameserver={{ env.bastion.networking.ip }} cio_ignore=all,!condev zfcp.allow_lun_scan=0 rd.znet=qeth,{{ node.networking.device1 }},{{ node.networking.device2 }},{{ node.networking.device3 }},layer2=1 {{ rd_zfcp_string }}'
```
AmadeusPodvratnik, line 70: There is no support for ipv6 (which is available for KVM) is this as designed? If yes, when do you plan to add ipv6 support?

inventories/default/host_vars/KVMhostname1-here.yaml.template
```diff
@@ -81,4 +81,10 @@ lpar:
 #      - 
 #      - 
 #    dev_num: 
-#    lun_name: 
\ No newline at end of file
+#    lun_name:
+
+# Section 7 -live disk info
```
AmadeusPodvratnik, line 86: This new section is for lpar and therefor you should change the description in line 16 to 19.

inventories/default/host_vars/KVMhostname2-here.yaml.template
```diff
@@ -81,4 +81,10 @@ lpar:
 #      - 
 #      - 
 #    dev_num: 
-#    lun_name: 
\ No newline at end of file
+#    lun_name:
```
AmadeusPodvratnik, line 84: This new section is for lpar and therefor you should change the description in line 16 to 19.

inventories/default/host_vars/KVMhostname3-here.yaml.template
```diff
@@ -81,4 +81,10 @@ lpar:
 #      - 
 #      - 
 #    dev_num: 
-#    lun_name: 
\ No newline at end of file
+#    lun_name:
```
AmadeusPodvratnik, line 84: This new section is for lpar and therefor you should change the description in line 16 to 19.

AmadeusPodvratnik (changes requested):
Hi @sandisamp Thx for the lpar. Pls add the description of the lpar to section 3 (Step 3: Set Variables (host_vars)) for the time being. 
In addition pls add a section to the Readme.md (New features) and add your lpar support to it.

inventories/default/host_vars/KVMhostname1-here.yaml.template
sandisamp: Updated in lines 16-19

roles/boot_LPAR/tasks/main.yaml
sandisamp: This will be added later. will conduct test and update on this.

roles/install_packages/tasks/main.yaml
sandisamp: Yes you are right, I have updated the code for it to only be installed in bastion and lpar based installation only.

roles/boot_LPAR/tasks/main.yaml
AmadeusPodvratnik, line 47: The subnet mast is hardcoded to 255.255.0.0. Better would be to use {{ node.networking.subnetmask }} instead.
In addition there is no ipv6 section. Pls add it to have ipv6 support.

roles/boot_LPAR/tasks/main.yaml
AmadeusPodvratnik, line 70: The subnet mast is hardcoded to 255.255.0.0. Better would be to use {{ node.networking.subnetmask }} instead.
In addition there is no ipv6 section. Pls add it to have ipv6 support.

playbooks/6_create_nodes.yaml
```diff
@@ -1,33 +1,71 @@
...
+  vars_files:
+    - "{{ inventory_dir }}/group_vars/all.yaml"
+  tasks:
+    - name: Loop over node types and include role
+      include_role:
+        name: check_for_lpar_nodes
+      loop: 
+        - bootstrap
+        - compute
+        - control
```
AmadeusPodvratnik, line 14: Do you want to not support the infra nodes than you should mention it in the description.

playbooks/5_setup_bastion.yaml
```diff
@@ -153,10 +161,11 @@
         - client.crt
         - client.key
         - ta.key
-      when: env.setup_openvpn == True and env.z.high_availability == True
+
+      when: env.setup_openvpn == True and env.z.high_availability == True  and env.bastion.installation_type != "lpar"
   roles:
-    - { role: robertdebock.epel, tags: openvpn, when: env.setup_openvpn == True and env.z.high_availability == True }
-    - { role: robertdebock.openvpn, tags: openvpn, when: env.setup_openvpn == True and env.z.high_availability == True }
+    - { role: robertdebock.epel, tags: openvpn, when: env.setup_openvpn == True and env.z.high_availability == True  and env.bastion.installation_type != "lpar"  }
```
AmadeusPodvratnik, line 167: Pls use for all cases "env.bastion.installation_type|lower" for installation type because of the control variable "lpar". This would be more failure tolerant and prevent issues in case of typos or camelcase values.

AmadeusPodvratnik (changes requested):
@sandisamp Thank you for the PR. Pls see my comments.

roles/dns/tasks/main.yaml
```diff
@@ -135,6 +135,22 @@
 
 - name: Restart network to update changes made to /etc/resolv.conf
   tags: dns, resolv
-  ansible.builtin.service:
-    name: NetworkManager
-    state: restarted
+  block:
```
smolin-de, line 138: This patch is not required anymore.
We use RHEL 9 as default OS.

roles/dns/tasks/initial-resolv.yaml
```diff
@@ -18,7 +18,22 @@
     mode: "644"
 
 - name: Restart network to update changes made to /etc/resolv.conf
-  tags: resolv
-  ansible.builtin.service:
-    name: NetworkManager
-    state: restarted
+  tags: dns, resolv
```
smolin-de, line 21: This patch is not required anymore.
We use RHEL 9 as default OS.

docs/set-variables-host-vars.md
```diff
@@ -81,4 +87,11 @@
 **lpar.storage_group_2.type** | <b>(Optional)</b> Storage type. FCP is the only tested type as of now. | fcp
 **lpar.storage_group_2_.storage_wwpn** | <b>(Optional)</b> World-wide port numbers for storage group. Use provided list formatting. | 500708680235c3f0<br />500708680235c3f1<br />500708680235c3f2<br />500708680235c3f3
 **lpar.storage_group_2_.dev_num** | <b>(Optional)</b> The logical device number of the Host Bus Adapter (HBA) for the storage group. | C001
-**lpar.storage_group_2_.lun_name** | <b>(Optional)</b> he Logical Unit Numbers (LUN) that points to a specific virtual disk<br /> behind the WWPN. | 4200569309ahhd240000000000000c001
\ No newline at end of file
+**lpar.storage_group_2_.lun_name** | <b>(Optional)</b> he Logical Unit Numbers (LUN) that points to a specific virtual disk<br /> behind the WWPN. | 4200569309ahhd240000000000000c001
+
+## 7 - (Optional) Livedisk info
+**Variable Name** | **Description** | **Example**
```
smolin-de, line 93: Where do I found some information about live disk ?
How do I create such live disk created ?
It is defined as optional, but I guess it is mandatory  when I want to use LPAR install. Correct ?

playbooks/0_setup.yaml
```diff
@@ -39,7 +39,7 @@
       tags: lpar_check
       stat:
         path: "{{ inventory_dir }}/host_vars/{{ env.z.lpar1.hostname }}.yaml"
-      when: env.z.lpar1.hostname is defined
+      when: env.z.lpar1.create == True
```
smolin-de, line 42: Why is this changed ?

playbooks/5_setup_bastion.yaml
```diff
@@ -155,10 +163,10 @@
         - client.crt
         - client.key
         - ta.key
-      when: env.setup_openvpn == True and env.z.high_availability == True
+      when: env.z.high_availability == True and env.bastion.installation_type|lower != "zvm" and env.bastion.installation_type|lower != "lpar"
   roles:
```
smolin-de, line 167: why is" env.setup_openvpn == True" removed from the "when" check ?

playbooks/6_create_nodes.yaml
smolin-de, line 14: I do not see an error check ?
What happen if bootstrap install fails ?
Will the creation of compute nodes and control nodes continue ?

playbooks/6_create_nodes.yaml
```diff
@@ -64,20 +102,54 @@
...
       ignore_errors: true
+      when: env.cluster.nodes.bootstrap.vm_name not in hosts_with_host_vars
 
     - name: Undefine bootstrap. Expect ignored errors if bootstrap is already undefined.
       tags: create_nodes, teardown_bootstrap
       community.libvirt.virt:
         name: "{{ env.cluster.nodes.bootstrap.vm_name }}"
         command: undefine
       ignore_errors: true
+<<<<<<< HEAD
```
smolin-de, line 113: Bad patch !
Merge conflicts are not resolved !!!!!

playbooks/6_create_nodes.yaml
```diff
@@ -64,20 +102,54 @@
...
+    - name: Restart the Haproxy Service
+      tags:  create_nodes, teardown_bootstrap
+      systemd_service:
+        name: haproxy
+        state: restarted
+      when: env.cluster.nodes.bootstrap.vm_name not in hosts_with_host_vars
+      
+=======
 
+>>>>>>> 077c676 (Revert "Deleting entries related to Bootstrap in Haproxy post its del… (#255))
```
smolin-de, line 133: Bad patch !!
Resolve merge conflicts correctly.

docs/set-variables-host-vars.md
sandisamp: See under important note in same file. It is marked as optional because this file is same for creating regular LPARs as well as LPAR info for booting lpar as a node. The initial use case was just for defining LPAR as a host for KVM nodes.

playbooks/6_create_nodes.yaml
```diff
@@ -1,33 +1,71 @@
...
+- name: 6 create nodes - find lpar host files
+  hosts: localhost
+  vars_files:
+    - "{{ inventory_dir }}/group_vars/all.yaml"
+  tasks:
+    - name: Loop over node types and include role
+      include_role:
+        name: check_for_lpar_nodes
+      loop: 
+        - bootstrap
```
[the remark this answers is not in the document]
AmadeusPodvratnik: Maybe as am option but not necessary. If you have only lpars than it make sense to use them even for bootstraping. Turning bootstrap node to compute node would make sense. 
We are doing this in several projects. Not sure how simple this to be integrated otherwise another PR makes sense.

roles/dns/tasks/initial-resolv.yaml
sandisamp: I think this patch can remain as it works both with rhel8.x and rhel9.x

playbooks/5_setup_bastion.yaml
sandisamp: Added the check

playbooks/6_create_nodes.yaml
sandisamp: If there is a boot error the Hmc python script will detect that and will result in FAIL. However configuration errors might need to be debugged and are currently not detectable on boot.

playbooks/6_create_nodes.yaml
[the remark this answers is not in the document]
sandisamp: I have added support to use bastion and bootstrap to be KVM, with this there is a possibility of not underutilising lpar resources.

For converting bootstrap to worker we can add in a later patch.

roles/dns/tasks/initial-resolv.yaml
smolin-de: RHEL8 is not supported anymore !!!

playbooks/5_setup_bastion.yaml
```diff
@@ -82,15 +82,23 @@
     - { role: sno_haproxy, when: env.bastion.options.loadbalancer.on_bastion and control_node_count | int == 1 }
     - httpd
 
+- name: Install tessia-baselib for lpar based installtions
+  hosts: bastion
+  tags: services, section_2
+  vars_files: 
+  - "{{ inventory_dir }}/group_vars/all.yaml"
+  roles:
+  - { role: install_tessia_baselib, when: env.installation_type|lower == "lpar" }
```
smolin-de, line 91: Bastion user is not necessary a root user. Can a  user, e.g "admin", also install the tessia library ?

smolin-de: If not, then you have duplicated code. the lines 85 to 90 should be deleted.

playbooks/6_create_nodes.yaml
```diff
@@ -64,20 +102,34 @@
...
+      when: env.cluster.nodes.bootstrap.vm_name not in hosts_with_host_vars
 
     - name: Undefine bootstrap. Expect ignored errors if bootstrap is already undefined.
       tags: create_nodes, teardown_bootstrap
       community.libvirt.virt:
         name: "{{ env.cluster.nodes.bootstrap.vm_name }}"
         command: undefine
       ignore_errors: true
-
+
```
smolin-de, line 113: In KVM mode we delete / destroy the bootstrap node.
What happen with the bootstrap node in LPAR mode. Is it still running ?

playbooks/6_create_nodes.yaml
```diff
@@ -64,20 +102,34 @@
...
+  tasks:
+    - name: boot compute nodes
+      vars:
+        node_type: "compute"
+        node_name: "{{ item }}"
+        ignition: "worker"
+      include_tasks: 
+        file: ../roles/boot_LPAR/tasks/main.yaml
+      loop: "{{ q('list',env.cluster.nodes[node_type].vm_name) | flatten }}"
+      when: item in hosts_with_host_vars
```
smolin-de, line 135: This function requires a root user or a user with root access.
Consider to add the line "become: true", in case the user has defined a non root user on his bastion host.

smolin-de (commented):
I am not able to test the code, but it seems to me that this code does not work if the user defines a non-root user for his bastion host.

playbooks/6_create_nodes.yaml
sandisamp: Yes it will not be teared down or shutdown, as per my previous comment, will add a patch to convert the bootstrap to worker node.

playbooks/5_setup_bastion.yaml
sandisamp: Tessie baselib installation is only called once in the playbook from here, other instance is for hypershift. 

So removing this will break the liar installation as no Tessia baselib will be present on the bastion

playbooks/6_create_nodes.yaml
sandisamp: Root user is not required to boot lpar nodes, only the python path with the tessia baselib should be accessible. The credentials for HMC will be picked from the configuration. These would suffice to boot.

playbooks/5_setup_bastion.yaml
smolin-de: I did not said to remove all lines. The line 91 should be kept. The lines 85 to 90 could be deleted, if you can install the code with a non-root user.

playbooks/6_create_nodes.yaml
smolin-de: When you convert the bootstrap node to a worker node in a later step, what kind of DNS name will use use. Can you give me an example ?

playbooks/6_create_nodes.yaml
smolin-de: The boot lpar script will be copied to: "/root/ansible_workdir/boot_lpar.py". This is a hardcoded value. Only root or sudo users have a access to it. Did you test your code with an not-root bastion user ?

playbooks/6_create_nodes.yaml
sandisamp: This node should resolve to the bootstrap hostname entry in the dns, and once completed the haproxy would need to be updated.

playbooks/6_create_nodes.yaml
AmadeusPodvratnik: My view on this is, that we can live with that because we have other places doing more or less the same. So for the time being it is ok. We need to convert bootstrap to worker node like we do it in other automation tools.

playbooks/6_create_nodes.yaml
AmadeusPodvratnik: This is ok for me.

AmadeusPodvratnik (approved):

PR #244, opened 2024-03-08
fix: Remove pipefail option

smolin-de:
Remove pipefail option and simplify loop definition.


AmadeusPodvratnik (approved):
Thank you @smolin-de .. looks good so far.

PR #245, opened 2024-03-11
feat: Add support to create a three node OCP cluster

smolin-de:
With this patch you can create a three node OCP cluster, without dedicated compute nodes. You just need to uncomment the compute node hostname definition in the config file (all.yaml).

### DIFF
--- a/docs/run-the-playbooks.md
+++ b/docs/run-the-playbooks.md
@@ -19,11 +19,19 @@ ansible-playbook playbooks/0_setup.yaml
 * Watch Ansible as it completes the installation, correcting errors if they arise.
 * To look at what tasks are running in detail, open the playbook or roles/role-name/tasks/main.yaml
 * Alternatively, to run all the playbooks at once, start the master playbook by running this shell command:
-```
+
+```shell
 ansible-playbook playbooks/site.yaml
 ```
+
 * If the process fails in error, go through the steps in the [troubleshooting](troubleshooting.md) page.
 * At the end of the the last playbook, follow the printed instructions for first-time login to the cluster.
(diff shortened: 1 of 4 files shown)


mohammedzee1000:
This is pretty cool. 
Although one questions, what do you mean by 'uncomment the compute node host name definition'. Isnt compute node section uncommented by default

https://github.com/IBM/Ansible-OpenShift-Provisioning/blob/b440fe5d4380199713ed10b7c36fa96a768afa69/inventories/default/group_vars/all.yaml.template#L135-L152

smolin-de:
> This is pretty cool. Although one questions, what do you mean by 'uncomment the compute node host name definition'? Isn't compute node section uncommented by default?
> 
> https://github.com/IBM/Ansible-OpenShift-Provisioning/blob/b440fe5d4380199713ed10b7c36fa96a768afa69/inventories/default/group_vars/all.yaml.template#L135-L152

You are right. We do not need to uncomment compute node definition, when my patch is applied. It is the default value. I forgot it, because with the current code you must define compute nodes.

AmadeusPodvratnik (commented):
Hi @smolin-de , thank you for this PR. Very cool and one step closer to SNO. Nevertheless I have a question regarding different AOP runs. For example someone is testing your 3 Node cluster and next it would like to test a 5 node cluster (or vice versa). How the correct haproxy settings will be exchanged? What steps the user need to be followed?
Due to the fact that this is an 'implicit' feature only handled by setting some variables, I would suggest to document this feature. Maybe it's worth to start a table documenting the features we support (similar we track internally using mural).

AmadeusPodvratnik (approved):
@smolin-de Thank you.

Issue #247, opened 2024-03-14
env.cluster.nodes.compute.mac documented but not exists in all.yaml.template 


AmadeusPodvratnik:
Variable env.cluster.nodes.compute.mac is documented but missing in all.yaml.template.

PR #248, opened 2024-03-14
fix: Mac specification documented but not in all.yaml.template.

AmadeusPodvratnik:
fix: Fix issue #247. Missing mac variable in all.yaml.template. 
Variable was documented but did not exist in the template.


veera-damisetti (approved):
Hi @AmadeusPodvratnik 
Looks good

PR #249, opened 2024-03-14
Deleting entries related to Bootstrap in Haproxy post its deletion

k-shiva-sai:
Deleteing all the entries related bootstrap  in haproxy config after its    ###deletion. # # #


AmadeusPodvratnik (approved):
@k-shiva-sai Thanks. Looks good.

PR #250, opened 2024-03-15
fix: Disable UDP aggregation on additional compute nodes

smolin-de:
Currently, UDP aggregation is not supported on IBM Z and is not automatically deactivated on multi-architecture compute clusters with an x86_64 control plane and additional s390x compute machines. To ensure that the addtional compute nodes are added to the cluster correctly, you must disable UDP aggregation.


roles/update_ignition_files/tasks/main.yml
```diff
@@ -1,5 +1,29 @@
 ---
 
+# Patch cluster configuration: Disable UDP aggregation
+# https://docs.openshift.com/container-platform/4.15/post_installation_configuration/configuring-multi-arch-compute-machines/creating-multi-arch-compute-nodes-ibm-z-kvm.html#machine-user-infra-machines-ibm-z-kvm_creating-multi-arch-compute-nodes-ibm-z-kvm
```
AmadeusPodvratnik, line 4: You just disable UDP aggregation without a check if current installed cluster is on s390x or contains s390x parts.
As far as I understood it's possible to start e.g. a cluster on x86 (master nodes) and add later additional nodes to the cluster. The additional nodes could be still x86 and therefor disable of the UDP aggregation will not be required, correct? Is it possible to get knowledge  about the current arch of the existing nodes?
What is about arm?

AmadeusPodvratnik (commented):
@smolin-de Thank you for the PR. Pls see my comments.

roles/update_ignition_files/tasks/main.yml
smolin-de: @AmadeusPodvratnik 
I resolved the issues.

AmadeusPodvratnik (approved):
@smolin-de Thx.

PR #251, opened 2024-03-15
fix: Enable HAProxy statistics report port 1936

smolin-de:
Port 1936 is used to get HAProxy statistics report via web browser.


AmadeusPodvratnik (approved):
@smolin-de Thank you for the PR.

PR #253, opened 2024-03-21
fix: DNS issues

### LINKED ISSUE #252: setup bastion fails for 3 node cluster definition
Playbook 5 fails for a 3 node cluster:

TASK [dns : Add compute nodes to DNS reverse lookup file on bastion] ************************************************************************************************************
fatal: [bastion]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'None' has no attribute 'split'. 'None' has no attribute 'split'\n\nThe error appears to be in '/Users/pod/Documents/Business/OCP/DevItems/release/Ansible-OpenShift-Provisioning/roles/dns/tasks/main.yaml': line 97, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Add compute nodes to DNS reverse lookup file on bastion\n  ^ here\n"}

# Section 9 - Compute Nodes
      compute:
        disk_size: 120
        ram: 16384
        vcpu: 4
        vcpu_model_option: "--cpu host"
        vm_name:
          - #X
          - #X
        ip:
          - #X
          - #X
        ipv6:
          - #X
          - #X
        mac:
          - #X
          - #X
        hostname:
          - #X
          - #X

smolin-de:
This patch fixes #252 'setup bastion fails for 3 node cluster'

### DIFF
--- a/roles/dns/tasks/main.yaml
+++ b/roles/dns/tasks/main.yaml
@@ -104,7 +104,7 @@
   loop_control:
     extended: yes
     index_var: i
-  when: env.cluster.nodes.compute.hostname is defined
+  when: env.cluster.nodes.compute.hostname is defined and env.cluster.nodes.compute.hostname[0] is defined and env.cluster.nodes.compute.hostname[0] != None
 
 - name: Add infrastructure nodes to DNS reverse lookup file on bastion
   tags: dns
--- a/roles/dns/tasks/main.yaml
+++ b/roles/dns/tasks/main.yaml
@@ -104,6 +104,7 @@
   loop_control:
     extended: yes
     index_var: i
+  when: env.cluster.nodes.compute.hostname is defined
 
 - name: Add infrastructure nodes to DNS reverse lookup file on bastion
   tags: dns
--- a/roles/check_dns/tasks/main.yaml
+++ b/roles/check_dns/tasks/main.yaml
@@ -52,7 +52,7 @@
   loop_control:
     extended: yes
     index_var: i
-  when: env.cluster.nodes.compute.hostname is defined
+  when: env.cluster.nodes.compute.hostname is defined and env.cluster.nodes.compute.hostname[0] is defined and env.cluster.nodes.compute.hostname[0] != None
 
 - name: Check infrastructure nodes DNS resolution
   tags: check_dns, dns
--- a/roles/check_dns/tasks/main.yaml
+++ b/roles/check_dns/tasks/main.yaml
@@ -52,6 +52,7 @@
   loop_control:
     extended: yes
     index_var: i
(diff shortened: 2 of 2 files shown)


roles/dns/tasks/main.yaml
```diff
@@ -104,6 +104,7 @@
   loop_control:
     extended: yes
     index_var: i
+  when: env.cluster.nodes.compute.hostname is defined
```
AmadeusPodvratnik, line 107: I would suggest to change to 
 when: env.cluster.nodes.compute.hostname[0] is defined and env.cluster.nodes.compute.hostname[0] != None 
 because our template has a defined compute node structure. We can't expect Ansible knowledge from user side so it is better to be more tolerant. 
 The same for the check_dns role.

AmadeusPodvratnik (changes requested):
Hi @smolin-de thx for the fix. Pls see my comments.

roles/dns/tasks/main.yaml
smolin-de: It is sufficient to check only for "env.cluster.nodes.compute.hostname[0] is defined".

I used the same check as we use already for infra nodes. "when: env.cluster.nodes.infra.hostname is defined" is used multiple time.

But, why we should allow the user to create empty variables ?
We should change our template and not allow empty variables.

AmadeusPodvratnik: Ok ... understood ... so how do you want to change the template? In addition we need to describe what to do for what kind of cluster, correct?

AmadeusPodvratnik: If template must be changed or the user need to do some specific steps than we need to document it. I personally would suggest to make the when clauses more error tolerant but can go with both ways ... document it or change the when clause.

smolin-de:
@AmadeusPodvratnik 
I updated this patch. 
I will create additional patches for other check improvements.

roles/dns/tasks/main.yaml
```diff
@@ -104,6 +104,7 @@
   loop_control:
     extended: yes
     index_var: i
+  when: env.cluster.nodes.compute.hostname is defined and env.cluster.nodes.compute.hostname[0] is defined and env.cluster.nodes.compute.hostname[0] != None
```
AmadeusPodvratnik, line 107: Line 62 contains 
when: env.cluster.nodes.compute.hostname is defined
but should have the same when statement.

AmadeusPodvratnik (changes requested):
Thank you @smolin-de , I identified another line to be changed.

AmadeusPodvratnik (approved):
@smolin-de Thank you.. looks good.

PR #256, opened 2024-03-22
fix: Remove empty array values in template file

smolin-de:


AmadeusPodvratnik (approved):
@smolin-de Thank you. Looks good.

PR #257, opened 2024-03-22
fix: Partial fix for #160 (RHEL support).

AmadeusPodvratnik:
Disable openvpn and set to deprecated and will be removed in upcoming releases.


veera-damisetti (approved):
Looks good

PR #259, opened 2024-03-25
fix: Missing day-2 variable definitions

smolin-de:


AmadeusPodvratnik (approved):
THx @smolin-de , looks good.

PR #265, opened 2024-04-03
perf: Disable memballon driver on KVM control and compute nodes

smolin-de:
Exclude the memory balloon device, because this is recommended in the OCP docs (chapter: "Recommended host practices for IBM Z & IBM LinuxONE environments".


AmadeusPodvratnik (approved):
Thx @smolin-de looks good so far.

PR #272, opened 2024-04-18
feat: Add RHEL9 support

smolin-de:
Add RHEL9 kickstart for bastion host (#160).
Support for openvpn must be removed in an additional PR, because the robertdebock.openvpn packages can not be installed on RHEL9.

### DIFF
--- a/roles/create_bastion/templates/bastion-ks.cfg.j2
+++ b/roles/create_bastion/templates/bastion-ks.cfg.j2
@@ -1,14 +1,13 @@
 # Template for bastion kickstart configuration file. Some parts come from the create_bastion role.
-# This kickstart file was tested with RHEL 8.7
-
-%pre --log=/root/pre.log
-%end
+# This kickstart file was tested with RHEL 9.2 and 9.3
 
 # Reboot after installation
 reboot
 
 # Use text mode install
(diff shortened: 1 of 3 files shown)


mohammedzee1000:
Will this also work if user has rhel 8 bastion?.

smolin-de:
> Will this also work if user has rhel 8 bastion?.

I did not test it with rhel 8.

mohammedzee1000:
ok just checking. It might be good to validate this. We dont know how quickly people will shift to rhel 9 bastion.

mohammedzee1000:
> > ok just checking. It might be good to validate this. We dont know how quickly people will shift to rhel 9 bastion.
> 
> But, we know that there will be issues. e.g. openvpn can not be installed on rhel9, some package name changed, the default python3 version will change from 3.6 to 3.9, ...

Might be good to keep both ks files then and pick based on rhel type? 
wdyt @AmadeusPodvratnik

smolin-de:
If agreed, then we should close the PR and use this PR instead: https://github.com/IBM/Ansible-OpenShift-Provisioning/pull/274

smolin-de:
We will use the PR #274

PR #273, opened 2024-04-19
fix: hostname does not contain a FQDN

smolin-de:


AmadeusPodvratnik (approved):
@smolin-de Thx for the PR. Looks good so far.

PR #274, opened 2024-04-23
feat: Add RHEL9 support

smolin-de:
Add RHEL9 kickstart for bastion host (#160).
Support for openvpn must be removed in an additional PR, because the robertdebock.openvpn packages can not be installed on RHEL9.

### DIFF
--- a/roles/create_bastion/tasks/main.yaml
+++ b/roles/create_bastion/tasks/main.yaml
@@ -7,11 +7,22 @@
     mode: "0755"
     state: directory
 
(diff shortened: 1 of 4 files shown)


smolin-de:
@mohammedzee1000 @AmadeusPodvratnik 
Here is a patch with RHEL8 and RHEL9, if you like it more. Use it.

AmadeusPodvratnik (approved):
@smolin-de Thx for the patch. Looks good so far.

PR #281, opened 2024-05-30
Added support for Hipersockets - zVM HCP

veera-damisetti:
- Added support for Hipersockets - zVM HCP
- Updated hcp creation command to accept additional flags from user
- Updated docs

### DIFF
--- a/inventories/default/group_vars/all.yaml.template
+++ b/inventories/default/group_vars/all.yaml.template
@@ -476,6 +476,7 @@ day2_compute_node:
   vm_mac:
   vm_interface:
   hostname:
+  host_user:
   host_arch:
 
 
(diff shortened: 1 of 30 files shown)


inventories/default/group_vars/all.yaml.template
```diff
@@ -297,6 +297,7 @@ hypershift:
     ocp_release:
     machine_cidr: 192.168.122.0/24
     arch:
+    additional_flags:
```
AmadeusPodvratnik, line 300: There is a new variable but not described what additional flags might be.

AmadeusPodvratnik (changes requested):
@veera-damisetti Thx for the PR. Looks good beside my comments on additional_flags.

PR #284, opened 2024-06-07
Added the Role for Reverse DNS when IPs of nodes falls under differen…

k-shiva-sai:
When the Bastion node IPs and OCP node IPs falls differnet subnet zones say 172.23.233.156 be bastion and 172.23.230.64 be master then while setting the rever DNS zone it assume nodes ips also falls under same zone and  adds the ip to bastion zone due to which nslookup fails. 
To allow this I have added a role multisubnet_reverse_dns which resolves like this.
```
$ORIGIN 230.23.172.in-addr.arpa.
64 IN PTR master0.ocpz.lnxne.boe.
65 IN PTR master1.ocpz.lnxne.boe.
66 IN PTR master2.ocpz.lnxne.boe.
54 IN PTR worker0.ocpz.lnxne.boe.
68 IN PTR worker1.ocpz.lnxne.boe.
72 IN PTR bootstrap.ocpz.lnxne.boe.
$ORIGIN 233.23.172.in-addr.arpa.
156 IN PTR bastion.ocpz.lnxne.boe.
156 IN PTR api.ocpz.lnxne.boe.
156 IN PTR api-int.ocpz.lnxne.boe.
```
and the named conf changes like this
```
zone "23.172.in-addr.arpa" IN {
     type master;
     file "/var/named/ocpz.rev";
     allow-update { any; };
     allow-query { any; };
};

```

### DIFF
--- a/inventories/default/group_vars/all.yaml.template
+++ b/inventories/default/group_vars/all.yaml.template
@@ -218,6 +218,7 @@ env:
 #    no:
 
 # Section 14 - (Optional) Misc
+  multi_subnet: False // Plese set this Option True for setting up the Reverse dns when IPs of nodes falls in differnet subnets
   language: en_US.UTF-8
   timezone: America/New_York
   keyboard: us
--- a/roles/boot_LPAR/templates/boot_lpar.py
+++ b/roles/boot_LPAR/templates/boot_lpar.py
@@ -78,6 +78,9 @@ lpar_parameters = {
         }
     }
 }
-
+if lpar_name[0:3] in ["m13", "m42"]:
+    lpar_parameters["boot_params"]["lun"]="4003402b00000000"
+    lpar_parameters["boot_params"]["wwpn"]="500507630a1b50a4"
+print(lpar_parameters["boot_params"])
 hmc.start(lpar_name, lpar_cpu, lpar_memory, lpar_parameters)
 hmc.logoff()
--- a/roles/multisubnet_reverse_dns/tasks/main.yml
+++ b/roles/multisubnet_reverse_dns/tasks/main.yml
@@ -1,4 +1,8 @@
 ---
+- name: Creating the ansible_workdir
+  file:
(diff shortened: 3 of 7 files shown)


roles/multisubnet_reverse_dns/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
+---
+- name: Creating the ansible_workdir
+  file:
+    path: /root/ansible_workdir/
```
AmadeusPodvratnik, line 4: This might fail if user is not root user. For temporary files use e.g. /tmp folder where usually everyone has access to.

inventories/default/group_vars/all.yaml.template
```diff
@@ -218,6 +218,7 @@ env:
 #    no:
 
 # Section 14 - (Optional) Misc
+  multi_subnet: False // Plese set this Option True for setting up the Reverse dns when IPs of nodes falls in differnet subnets
```
AmadeusPodvratnik, line 221: Pls document this new parameter in the docs section, too.

roles/boot_LPAR/templates/boot_lpar.py
```diff
@@ -78,6 +78,9 @@ def list_of_strings(arg):
         }
     }
 }
-
+if lpar_name[0:3] in ["m13", "m42"]:
```
AmadeusPodvratnik, line 81: This is hardcoded and might lead to errors if something changed. Pls use wars for that if needed.

AmadeusPodvratnik (changes requested):
Hi @k-shiva-sai Thank you for the PR. Pls check my comments.

roles/boot_LPAR/templates/boot_lpar.py
sandisamp: This should be taken as a variable in the host vars

PR #286, opened 2024-06-11
Livedisk with FCP changes

k-shiva-sai:
Enables to use the FCP live disk to boot the nodes.

### DIFF
--- a/docs/set-variables-host-vars.md
+++ b/docs/set-variables-host-vars.md
@@ -1,4 +1,4 @@
-i# Step 3: Set Variables (host_vars)
+# Step 3: Set Variables (host_vars)
 
 ## Overview
 * Similar to the group_vars file, the host_vars files for each LPAR (KVM host) must be filled in. 
--- a/docs/set-variables-host-vars.md
+++ b/docs/set-variables-host-vars.md
@@ -92,8 +92,8 @@
 ## 7 - (Optional) Livedisk info
 **Variable Name** | **Description** | **Example**
 :--- | :--- | :---
-**lpar.livedisk.livedisktype** | <b>(Optional)</b> Storage type. DASD and SCSI are tested types as of now. | dasd
-**lpar.livedisk.lun** | <b>(Optional)</b> The Lunid of the disk when the livedisktype is SCSI. | 4003402b00000000
-**lpar.livedisk.wwpn** | <b>(Optional)</b> World-wide port number when livedisktype is SCSI. | 500507630a1b50a4
+**lpar.livedisk.livedisktype** | <b>(Optional)</b> Storage type. DASD and SCSI are tested types as of now. | dasd/scsi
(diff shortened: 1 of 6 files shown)


roles/boot_LPAR/tasks/main.yaml
```diff
@@ -64,6 +74,8 @@
           --initrd {{ rhcos_download_url }}{{ rhcos_live_initrd }} \
           --livedisktype {{ node.lpar.livedisk.livedisktype }} \
           --devicenr {{ node.lpar.livedisk.devicenr }} \
+          --livedisklun {{ live_disk_lun }} \
+          --livediskwwpn {{ live_disk_wwpn }} \
```
sandisamp, line 78: This is duplicate code

sandisamp (changes requested):

sandisamp (approved):

docs/set-variables-host-vars.md
```diff
@@ -92,6 +92,8 @@
 ## 7 - (Optional) Livedisk info
 **Variable Name** | **Description** | **Example**
 :--- | :--- | :---
-**lpar.livedisk.livedisktype** | <b>(Optional)</b> Storage type. DASD is the only tested type as of now. | dasd
-**lpar.livedisk.devicenr** | <b>(Optional)</b> the device no of the DASD live disk | c6h1
+**lpar.livedisk.livedisktype** | <b>(Optional)</b> Storage type. DASD and SCSI are tested types as of now. | dasd
```
AmadeusPodvratnik, line 95: Now you are supporting DASD and FCP devices (Do you really mean SCSI and not FCP?) and the question is what exactly need to be defined. It would be good to be precise. If no lun or wwpn is defined than the devicenr is taken as DASD with the lun and wwpn it's for FCP devices.

roles/boot_LPAR/templates/boot_lpar.py
```diff
@@ -78,6 +80,9 @@ def list_of_strings(arg):
         }
     }
 }
+if args.livedisklun!="na" and args.livediskwwpn!="na":
```
AmadeusPodvratnik, line 83: There is no further logic check. If FCP is being specified but no lun or wwpn given or if DASD specified and lun and wwpn given. It would be good to have some checking for logic errors.

AmadeusPodvratnik (changes requested):
Hi @k-shiva-sai Thank you for the PR. Pls see my comments.

roles/boot_LPAR/templates/boot_lpar.py
k-shiva-sai: Thanks Amadeus for the comment.
I will work on this and raise the commit.

docs/set-variables-host-vars.md
k-shiva-sai: Node boot Works only when the livedisktype is scsi it is throwing errors when it is fcp.

AmadeusPodvratnik (approved):
Thx @k-shiva-sai Looks good so far.

PR #298, opened 2024-06-25
LPAR boot with dasd disks

k-shiva-sai:
Boot the LPAR with DASD disks.


AmadeusPodvratnik (approved):
@k-shiva-sai thank you for the PR. Looks good.l

PR #300, opened 2024-07-09
Enhancement: Moving HCP variables to a separate var file 

veera-damisetti:
Enhancement: Moving HCP variables to a separate var file 

- Introduced a new all.yaml.hcp.template for hcp vars
- Removed all hcp related vars from all.yaml
- Restructured HCP vars to make it more user friendly
- Added support for DASD disk for KVM guests
- Updated docs


inventories/default/group_vars/hcp.yaml.template
```diff
@@ -0,0 +1,141 @@
...
+    clusters_namespace:
+    hosted_cluster_name:
+    basedomain:
+    ocp_release_image:
+    arch: s390x
+    additional_flags: 
+    # Make sure to enclose pull_secret in 'single quotes'
+    pull_secret:
+
+  bastion_params:
```
AmadeusPodvratnik, line 42: Would it make sense to have a common bastion hosting all necessary stuff (dns, haproxy, etc).
Due to the split it might be useful not to have to many variable dubs.

AmadeusPodvratnik (changes requested):
Hi @veera-damisetti Thx for the PR. Pls see my comments and in addition could you pls provide a README update in the what's new section. Thx.

roles/install_tessia_baselib/tasks/main.yaml
```diff
@@ -4,7 +4,7 @@
     name:
       - "{{ item }}"
     state: present
-  loop: "{{ env.pkgs.zvm }}"
+  loop: "{{ hcp.pkgs.zvm }}"
```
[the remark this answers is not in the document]
veera-damisetti: Sure @AmadeusPodvratnik
make sense
Updated

inventories/default/group_vars/hcp.yaml.template
veera-damisetti: you mean common bastion for Hosted Cluster and Management OCP cluster ?

AmadeusPodvratnik: Ok ... as discussed via Slack for the time being hcp is having it's own bastion.

roles/install_tessia_baselib/tasks/main.yaml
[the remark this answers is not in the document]
AmadeusPodvratnik: As discussed via Slack we need here a followup to not duplicate variables for every component. So we need to have a follow-up here. For the time being HCP is more or less independent from a regular OCP installation but is still part of the project and should fit into.

AmadeusPodvratnik (approved):
As discussed via Slack go for now because HCP is some sort of independent.

veera-damisetti:
Thanks @AmadeusPodvratnik

PR #309, opened 2024-08-09
fix: Hung condition during update of /etc/hosts

smolin-de:
This patch solves a hung condition during update of /etc/hosts when
you use a container as control host.


AmadeusPodvratnik (approved):
Hi @smolin-de thx for the PR. To my knowledge ansible will not be forced to perform unsafe writes but has a fall back option. So looks good to me.

PR #313, opened 2024-08-19
Separation mirror vars

AmadeusPodvratnik:
feat: Move disconnected variables out of all.yaml for better usability.


AmadeusPodvratnik:
@mohammedzee1000 Could you pls have a look, too. Thx in advance.

veera-damisetti:
Looks good to me 
@mohammedzee1000 can you please add your view also here
thanks

mohammedzee1000:
Code wise looks good
But i have one doubt. Have we tried running this when someone does not create `disconnected.yaml`?
I would assume the `disconnected.enabled` checks in the `0_setup` playbook would fail in such case
@AmadeusPodvratnik

AmadeusPodvratnik:
> Code wise looks good But i have one doubt. Have we tried running this when someone does not create `disconnected.yaml`? I would assume the `disconnected.enabled` checks in the `0_setup` playbook would fail in such case @AmadeusPodvratnik

Hi @mohammedzee1000 , yes I made a mistake by adding a template instead of the yaml itself. Disconnected must be handled as default values (due to the pure nature how it's integrated). We can think about to have a variable inside of all.yaml but basically we do not need a template for disconnected.

inventories/default/group_vars/.gitignore
```diff
@@ -1,4 +1,5 @@
 /*
 !.gitignore
 !all.yaml.template
-!hcp.yaml.template
\ No newline at end of file
+!disconnected.yaml.template
```
mohammedzee1000, line 4: Is this required anymore?

mohammedzee1000 (commented):
@AmadeusPodvratnik

inventories/default/group_vars/.gitignore
AmadeusPodvratnik: Hi @mohammedzee1000 Thank you.. No should be removed.

mohammedzee1000 (approved):
lgtm

veera-damisetti (approved):
LGTM

PR #314, opened 2024-08-22
add os-variant flag

sanjayy-ibm:
os-variant flag is added in virt-install command for ISO based installation


veera-damisetti (approved):
Looks good @sanjayy-ibm 
thanks

PR #318, opened 2024-08-26
chore: Increase 'approve_certs' wait time to 45 min

smolin-de:
Approve certificate wait time is increased to 45min.


AmadeusPodvratnik (commented):
Hi @smolin-de Thank you for the PR. Just a question about these wait times. Make a configurable value sense?

smolin-de:
> Hi @smolin-de Thank you for the PR. Just a question about these wait times. Make a configurable value sense?

I do not think that an configurable value make sense.
The install will fail, if the time is to small.

AmadeusPodvratnik (approved):
@smolin-de Thank you ... /lgtm

PR #323, opened 2024-09-02
Fix for several issues

AmadeusPodvratnik:
Add some fixes around 3 node cluster and lpar support, dns setting, ip_forward, etc. Include 
-initrd-inject "/{{ kvm_host_home.stdout }}/{{ env.file_server.cfgs_dir }}/{{ env.bastion.networking.hostname }}/bastion-ks.cfg" \ in case pf create bastion which works not consistent and env.file_server.cfgs_dir should contain the absolute path.
Furthermore section 1 is being flatten.


veera-damisetti (approved):
Thanks @AmadeusPodvratnik 
Looks good from my side

Issue #329, opened 2024-09-16
Under some circumstances the execution of playbooks first after fresh clone of repo


AmadeusPodvratnik:
ansible-playbook playbooks/0_setup.yaml 
[WARNING]: Unable to parse /Users/pod/Documents/OCP/code/GitHub/Repositories/Ansible-OpenShift-Provisioning/inventories/default as an inventory source
[WARNING]: No inventory was parsed, only implicit localhost is available
[WARNING]: provided hosts list is empty, only localhost is available. Note that the implicit localhost does not match 'all'

PLAY [localhost] ****************************************************************************************************************************************************************************

TASK [Gathering Facts] **********************************************************************************************************************************************************************
ok: [localhost]
ERROR! an undefined variable was found when attempting to template the vars_files item '{{ inventory_dir }}/group_vars/disconnected.yaml'

The error appears to be in '/Users/pod/Documents/OCP/code/GitHub/Repositories/Ansible-OpenShift-Provisioning/playbooks/0_setup.yaml': line 9, column 7, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

  vars_files:
    - "{{ inventory_dir }}/group_vars/disconnected.yaml"
      ^ here
We could be wrong, but this one looks like it might be an issue with
missing quotes. Always quote template expression brackets when they
start a value. For instance:

    with_items:
      - {{ foo }}

Should be written as:

    with_items:
      - "{{ foo }}"
MBP-von-Amadeus:Ansible-OpenShift-Provisioning pod$ ansible-playbook playbooks/0_setup.yaml --extra-var inventory_dir=inventory/default
[WARNING]: Unable to parse /Users/pod/Documents/OCP/code/GitHub/Repositories/Ansible-OpenShift-Provisioning/inventories/default as an inventory source
[WARNING]: No inventory was parsed, only implicit localhost is available
[WARNING]: provided hosts list is empty, only localhost is available. Note that the implicit localhost does not match 'all'

PLAY [localhost] ****************************************************************************************************************************************************************************

TASK [Check disconnected variables if disconnected cluster will be installed.] **************************************************************************************************************
[...]

PR #330, opened 2024-09-16
Fix: Issue #329: Under specific circumstances the run of playbooks is failing.

AmadeusPodvratnik:
Without hosts file the run of playbooks is failing with unknown variables.


isumitsolanki (approved):
Look good to me 

cc: @AmadeusPodvratnik @veera-damisetti

PR #331, opened 2024-09-18
fix: LPAR boot with DASD disks updating parm file with /dev/dasda

k-shiva-sai:
Modifind the boot_lpar role so that `coreos.inst.install_dev` changes as per the lpar disk type


AmadeusPodvratnik (commented):
Hi @k-shiva-sai Thx for the PR. 
/lgtm

AmadeusPodvratnik (approved):

PR #341, opened 2024-10-17
feat: ISO boot support for KVM agents  - HCP 

veera-damisetti:
Added support for ISO boot support for KVM agents - HCP 
Updated docs


AmadeusPodvratnik (approved):

PR #345, opened 2024-10-29
Support for disconnected cluster installation using ABI on kvm

isumitsolanki:


mohammedzee1000:
Not directly related to this pr but I see you have a role called prepare_config which is only used for abi. Maybe at some point rename this so we know its for abi or perhaps we can figure out a way to generalize the mirroring operations better so we can reuse them everywhere instead of duplication.

We don't have to fix this now but food for thought.

isumitsolanki:
Hi @mohammedzee1000 , sure

AmadeusPodvratnik (approved):
Hi @isumitsolanki, thx for the PR. /lgtm

Issue #347, opened 2024-11-07
Creation of ABI cluster is getting failed on KVM (os version rhel9)


isumitsolanki:

isumitsolanki:
This issue has been resolved and PR got merged https://github.com/IBM/Ansible-OpenShift-Provisioning/pull/348 , hence closing this issue

PR #348, opened 2024-11-07
Creation of ABI cluster is getting failed on KVM (os version rhel9) #347

isumitsolanki:
https://github.com/IBM/Ansible-OpenShift-Provisioning/issues/347


AmadeusPodvratnik (approved):
Hi @isumitsolanki looks good so far ... will test it.

PR #353, opened 2024-11-20
Reverted minor bugs in boot params LPAR

sandisamp:
Some hardcoded value had been set in the boot script during a feat implementation for hipersocket. 

Also reverted the ip parameter where gateway ip was replaced by nameserver ip.


AmadeusPodvratnik (approved):
Hi @sandisamp , thx for the PR. /lgtm

PR #358, opened 2024-12-02
feat: Support for LPAR as compute nodes for HCP 

veera-damisetti:
Support for LPAR as compute nodes for HCP

1. Network types 
  - OSA 
  - RoCE

2. Disk types 
  - FCP 
  -  DASD 
  -  NVMe
  
 Updated Documentation


docs/run-the-playbooks-for-hcp.md
```diff
@@ -31,6 +31,10 @@ bastion_root_pw: '<password_you_want_to_keep_for_bastion>'
 api_server: '<api-server-url ot management cluster>:<port>' 
 user_name: '<username >'
 password: '<password >'
+
```
AmadeusPodvratnik, line 34: Hi @veera-damisetti, we have a hmc user/pwd in host_vars and introducing another place to handle user/pwd. I think it would be better to have such elementary settings at one place.

veera-damisetti: Thanks @AmadeusPodvratnik , for quick review.
In case of HCP , there is no need for user to fill/modify host_vars for LPARs, so as we need to encrypt the credentials , i moved them to secrets.yaml.

veera-damisetti:
Sorry for updating the documentation at last moment.
please have a look @AmadeusPodvratnik

AmadeusPodvratnik (approved):

Issue #366, opened 2024-12-13
OpenShift Client in not getting download from file_server 


isumitsolanki:
OpenShift Client in not present in expected directory on file_server due to that is it not getting download.

In existing implementation of disconnect OpenShift Client is getting download from internet. (https://mirror.openshift.com/pub/openshift-v4/s390x/clients/ocp) 

In case of disconnected you machine will not have internet connection.

isumitsolanki:
This issue has been resolved and PR got merged.

Issue #365, opened 2024-12-13
ABI: On fresh z/VM bastion ABI playbook is not installing tessia base lib


isumitsolanki:

isumitsolanki:
This issue has been fixed and PR got merged into main

PR #367, opened 2024-12-13
OpenShift Client issue has been resolved along with DNS issue

isumitsolanki:
This is pull request is having fix for below two issue  (ABI Disconnected)
[OpenShift Client in not getting download from file_server ](https://github.com/IBM/Ansible-OpenShift-Provisioning/issues/366)

[ABI: On fresh z/VM bastion ABI playbook is not installing tessia base lib](https://github.com/IBM/Ansible-OpenShift-Provisioning/issues/365)

### DIFF
--- a/roles/offline_artifacts/tasks/main.yaml
+++ b/roles/offline_artifacts/tasks/main.yaml
@@ -26,7 +26,6 @@
       loop:
         - "{{ env.file_server.protocol }}://{{ env.file_server.user + ':' + env.file_server.pass + '@' if env.file_server.protocol == 'ftp' else '' }}{{ env.file_server.ip }}{{ ':' + env.file_server.port if env.file_server.port | default('') | length > 0 else '' }}/{{ disconnected.mirroring.file_server.clients_dir }}/{{ disconnected.mirroring.file_server.oc_mirror_tgz }}"
         - "{{ disconnected.mirroring.client_download.ocp_download_url }}{{ disconnected.mirroring.client_download.ocp_client_tgz }}"
-        # - "{{ env.file_server.protocol }}://{{ env.file_server.user + ':' + env.file_server.pass + '@' if env.file_server.protocol == 'ftp' else '' }}{{ env.file_server.ip }}{{ ':' + env.file_server.port if env.file_server.port | default('') | length > 0 else '' }}/{{ disconnected.mirroring.file_server.clients_dir }}/{{ disconnected.mirroring.client_download.ocp_client_tgz }}"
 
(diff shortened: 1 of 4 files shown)


roles/offline_artifacts/tasks/main.yaml
```diff
@@ -25,7 +25,8 @@
...
-        - "{{ env.file_server.protocol }}://{{ env.file_server.user + ':' + env.file_server.pass + '@' if env.file_server.protocol == 'ftp' else '' }}{{ env.file_server.ip }}{{ ':' + env.file_server.port if env.file_server.port | default('') | length > 0 else '' }}/{{ disconnected.mirroring.file_server.clients_dir }}/{{ disconnected.mirroring.client_download.ocp_client_tgz }}"
+        - "{{ disconnected.mirroring.client_download.ocp_download_url }}{{ disconnected.mirroring.client_download.ocp_client_tgz }}"
+        # - "{{ env.file_server.protocol }}://{{ env.file_server.user + ':' + env.file_server.pass + '@' if env.file_server.protocol == 'ftp' else '' }}{{ env.file_server.ip }}{{ ':' + env.file_server.port if env.file_server.port | default('') | length > 0 else '' }}/{{ disconnected.mirroring.file_server.clients_dir }}/{{ disconnected.mirroring.client_download.ocp_client_tgz }}"
```
AmadeusPodvratnik, line 29: Hi @isumitsolanki, you commented the code. Is this your intention? 
If the code is not being used anymore, than delete it because this keeps the code clean. Thx.

AmadeusPodvratnik (changes requested):
Hi @isumitsolanki Thank you for the PR, pls see my comment.

roles/offline_artifacts/tasks/main.yaml
isumitsolanki: Hi @AmadeusPodvratnik , I just want to unblock solution team for that I have commented but this we have to discuss and then we will draw a solution. I  think whatever this written here the correct logic but solution team don't have access to keep oc client in the mentioned directory over filer-server. so in that case for as now i enabled to reach from internet as well.

AmadeusPodvratnik: This is understood but why you can't remove the lines and if needed bring it back?

isumitsolanki: ok sure @AmadeusPodvratnik , I will do the same

AmadeusPodvratnik (approved):
@isumitsolanki Thank you. /lgtm

PR #371, opened 2025-01-03
adapt to dpm LPARs

vasantkumar60:
Enables AOP to install on dpm LPAR clusters. 
Adding a flag 'dpm_enabled' in host vars. When true uuid is used for the livedisk, which has to be provided by the user.


sandisamp:
HI @vasantkumar60 thanks for raising the PR. 

As per our previous tests this was working both on the pr/sm and dpm mode for installing in LPAR.

What is the need for using the live disk uuid. Is it LPAR specific or any updates to the tessia baselib, that now require uuid as a parameter.
Kindly provide the justification for this PR.

Also since DPM lpars are already supported kindly add "fix" to the title, and make it more descriptive of the change.(refer other pull requests)

veera-damisetti:
Had a discussion with @vasantkumar60  offline , on changes required.

veera-damisetti (approved):
Thanks for the PR @vasantkumar60 
looks good to me

Issue #360, opened 2024-12-03
Playbook 3_setup_kvm_host is getting failed on LPAR os version rhel9 or above


isumitsolanki:
Playbook 3_setup_kvm_host is getting failed on LPAR os version rhel9 or above at the stage of package installation.

TASK [install_packages : Installing required packages for Linux machines.] ********************************************
ok: [m42lp53.lnxero1.boe] => (item=libguestfs)
ok: [m42lp53.lnxero1.boe] => (item=libvirt-client)
ok: [m42lp53.lnxero1.boe] => (item=libvirt-daemon-config-network)
ok: [m42lp53.lnxero1.boe] => (item=libvirt-daemon-kvm)
ok: [m42lp53.lnxero1.boe] => (item=cockpit-machines)
failed: [m42lp53.lnxero1.boe] (item=libvirt-devel) => {"ansible_loop_var": "item", "changed": false, "failures": ["No package libvirt-devel available."], "item": "libvirt-devel", "msg": "Failed to install some of the specified packages", "rc": 1, "results": []}
ok: [m42lp53.lnxero1.boe] => (item=virt-top)
ok: [m42lp53.lnxero1.boe] => (item=qemu-kvm)
ok: [m42lp53.lnxero1.boe] => (item=python3-lxml)
ok: [m42lp53.lnxero1.boe] => (item=cockpit)
ok: [m42lp53.lnxero1.boe] => (item=lvm2)
changed: [m42lp53.lnxero1.boe] => (item=httpd)

PLAY RECAP ************************************************************************************************************
127.0.0.1                  : ok=34   changed=12   unreachable=0    failed=0    skipped=44   rescued=0    ignored=0   
m42lp53.lnxero1.boe        : ok=3    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0

isumitsolanki:
This issue has been fixed and PR got merged into main

PR #373, opened 2025-01-08
Playbook 3_setup_kvm_host is getting failed on LPAR os version rhel9 or above

isumitsolanki:
RCA: Playbook was getting failed due to libvirt-devel packages that is not available for rhel9. 
Fix: Removed libvirt-devel from packages list and installing it only when rhel version in below 9.

This PR contain fix for [Playbook 3_setup_kvm_host is getting failed on LPAR os version rhel9 or above](https://github.com/IBM/Ansible-OpenShift-Provisioning/issues/360)  .


AmadeusPodvratnik (commented):
@isumitsolanki Thank you for the PR. /lgtm

AmadeusPodvratnik (approved):

PR #379, opened 2025-01-29
Included task to ssh into control and compute nodes post successful installation of cluster

sindhujabd268:
Included task to ssh into control and compute nodes post successful installation of cluster.


AmadeusPodvratnik (approved):
@sindhujabd268 Thank you for the PR. /lgtm

PR #380, opened 2025-02-04
Enhancement: Segregating hcp yaml playbook

jpattara:
Adding the 'Download Required images for booting Agents' and 'Configure httpd on bastion for hosting rootfs' steps on create agents yaml. 

So we can directly run the second playbook (create_agents_and_wait_for_install_complete.yaml) if hcp is already available to boot the compute nodes by downloading the required images.


veera-damisetti:
Thanks @jpattara  for the PR 
looks good to me

veera-damisetti (approved):

Issue #381, opened 2025-02-06
Bastion is not coming up fips enabled.


isumitsolanki:
Fips should be enabled on bastion if it's true.

isumitsolanki:
This issue has been resolved and PR git merged.

PR #382, opened 2025-02-06
Fix has been added for Bastion is not coming up fips enabled

isumitsolanki:
https://github.com/IBM/Ansible-OpenShift-Provisioning/issues/381


AmadeusPodvratnik (approved):
@isumitsolanki thanks for this PR. /lgtm

PR #395, opened 2025-03-17
Changes to enable IPSEC Network Configuration

suman-jainkeri:
This PR handles the changes to enable IPSEC Network Config changes during the installation.
- The boolean flag - **ipsec_enabled** - is added in the all.yaml to enable IPSEC configuration. Default value if False.
- Maintain the Network config manifest - **cluster-network-03-config.yml** - in the **files/** directory of **get_ocp** role.
- After the manifests are created, an additional Network manifest - **cluster-network-03-config.yml** - which has the configuration for enabling IPSEC is copied to the manifests directory, only when **ipsec_enabled flag is True**.

### DIFF
--- a/docs/set-variables-group-vars.md
+++ b/docs/set-variables-group-vars.md
@@ -201,8 +201,10 @@
 :--- | :--- | :---
 **abi.flag** | This is the flag, Will be used to identify during execution. Few checks in the playbook will be depend on this (default value will be False)  | True
 **abi.ansible_workdir** | This will be work directory name, it will keep required data that need to be present during or after execution | ansible_workdir
-**abi.ocp_installer_version** | Version will contain value of openshift-installer binary version user desired to be used | '4.15.0-rc.8'
-**abi.ocp_installer_url** | This is the base url of openshift installer binary it will remain same as static value, User Do not need to give value until user wants to change the mirror | 'https://mirror.openshift.com/pub/openshift-v4/s390x/clients/ocp/'
+**abi.ocp_installer_version** | Version will contain value of openshift-installer binary version user desired to be used | '4.15.8'
+**abi.ocp_installer_base_url** | This is the base url of openshift installer binary it will remain same as static value, User Do not need to give value until user wants to change the mirror | 'https://mirror.openshift.com/pub/openshift-v4/'
+**abi.architecture** | The installer binary supports two architecture options: multi and s390x. Users are required to specify the appropriate architecture value based on their deployment environment. | 'multi/s390x'
+**abi.boot_method** | Specifies the boot type for Agent-based Installation (ABI). Users must choose either iso or pxe based on the deployment method. Note: iso boot is supported only on KVM platforms. | 'iso/pxe'
 
 ## OpenShift Settings
 * The parameters bellow have a hierachical structure and need to be added to all.yaml in given format. For example if you want to change the hyperthreading (disable) than you need to specify the following value in all.yaml file:
--- a/docs/set-variables-group-vars.md
+++ b/docs/set-variables-group-vars.md
@@ -166,6 +166,7 @@
 **env.jumphost.user** | (Optional) The user name to login to the jumphost. | admin
 **env.jumphost.pass** | (Optional) The password for user to login to the jumphost. | ch4ngeMe!
 **env.jumphost.path_to_keypair** | (Optional) The absolute path to the public key file on the jumphost to be copied to the bastion. | /home/admin/.ssh/id_rsa.pub
+**env.ipsec_enabled** | If IPSEC network configuration has to be enabled, this flag should be set to True, default value is set to False. | False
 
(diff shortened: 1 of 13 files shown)


roles/get_ocp/files/cluster-network-03-config.yml
```diff
@@ -0,0 +1,8 @@
+apiVersion: operator.openshift.io/v1
+kind: Network
+metadata:
+  name: cluster
+spec:
+  defaultNetwork:
+    ovnKubernetesConfig:
+      ipsecConfig: {}
```
AmadeusPodvratnik, line 8: I do not see any content like mode for ipsecConfig so I'm wondering if there should be some additional code other than copy it into manifest directory.

AmadeusPodvratnik (commented):
Hi @suman-jainkeri thank you for your PR. Pls see my comment about cluster-network-03-config.yml file.

roles/get_ocp/files/cluster-network-03-config.yml
suman-jainkeri: Hi @AmadeusPodvratnik,
I will check and update the changes. Thank you for reviewing.

suman-jainkeri: Hi @AmadeusPodvratnik ,

I verified IPSec with configuring ipsec mode as **Full**
Briefly, **Full** mode is to encrypt pod-to-pod traffic and, optionally, traffic to external hosts.
Cluster came up and verified if IPSEC is enabled 

`[root@t313lp32 ~]# oc get pods -n openshift-ovn-kubernetes -l=app=ovnkube-node
NAME                 READY   STATUS    RESTARTS   AGE
ovnkube-node-2vgkm   8/8     Running   0          11m
ovnkube-node-2vxdr   8/8     Running   8          53m
ovnkube-node-4lvhv   8/8     Running   8          53m
ovnkube-node-chhkp   8/8     Running   0          18m
ovnkube-node-g6w2g   8/8     Running   8          53m
[root@t313lp32 ~]# 
[root@t313lp32 ~]# 
[root@t313lp32 ~]# oc -n openshift-ovn-kubernetes rsh ovnkube-node-2vgkm ovn-nbctl --no-leader-only get nb_global . ipsec
Defaulted container "ovn-controller" out of: ovn-controller, ovn-acl-logging, kube-rbac-proxy-node, kube-rbac-proxy-ovn-metrics, northd, nbdb, sbdb, ovnkube-controller, kubecfg-setup (init)
true
[root@t313lp32 ~]# `

With the above verification procedure, you can verify that IPsec is enabled between pods on your cluster when IPsec is configured in Full mode.

docs/set-variables-group-vars.md
```diff
@@ -166,6 +166,7 @@
 **env.jumphost.user** | (Optional) The user name to login to the jumphost. | admin
 **env.jumphost.pass** | (Optional) The password for user to login to the jumphost. | ch4ngeMe!
 **env.jumphost.path_to_keypair** | (Optional) The absolute path to the public key file on the jumphost to be copied to the bastion. | /home/admin/.ssh/id_rsa.pub
+**env.ipsec_enabled** | If IPSEC network configuration has to be enabled, this flag should be set to True, default value is set to False. | False
```
AmadeusPodvratnik, line 169: In general we should reduce the amount of vars in all.yaml so the default should not be required to be specified. So default should be false (set to False or no value specified).

inventories/default/group_vars/all.yaml.template
```diff
@@ -200,6 +200,7 @@ env:
   use_ipv6: True
   use_dhcp: False
   setup_openvpn: False
+  ipsec_enabled: False
```
AmadeusPodvratnik, line 203: Pls comment this var out (# ipsec_enabled: False) so that a user has an example how to set this.

roles/get_ocp/tasks/main.yaml
```diff
@@ -92,6 +92,24 @@
     /root/ocpinst/openshift-install create manifests --dir=/root/ocpinst/
   become: true
 
+- name: Copy the file when ipsec flag is enabled
+  tags: get_ocp
+  become: true
+  copy:
+    src: cluster-network-03-config.yml
+    dest: /root/ocpinst/manifests/cluster-network-03-config.yml
+  when: env.ipsec_enabled
```
AmadeusPodvratnik, line 101: Pls adapt the when clause and add "env.ipsec_enabled is defined" to enable a more lazy way to handle variables and minimize the change in all.yaml.

AmadeusPodvratnik (changes requested):

AmadeusPodvratnik:
@suman-jainkeri Thank you for the PR. Pls see my comments. Thank you.

AmadeusPodvratnik (approved):
Hi @suman-jainkeri Thank you. The PR looks good.

PR #403, opened 2025-04-21
Fix OCP client/installer URL logic to support multi and s390x architecture

Neeraj8418:
Updated main.yaml to dynamically construct the correct download URLs based on the architecture specified in all.yaml. This ensures proper support for both 'multi' and 's390x' when downloading OpenShift installer and client binaries.

### DIFF
--- a/inventories/default/group_vars/all.yaml.template
+++ b/inventories/default/group_vars/all.yaml.template
@@ -229,8 +229,9 @@ day2_compute_node:
 abi:
   flag: False
   ansible_workdir: 'ansible_workdir'
-  ocp_installer_version: '4.15.0-rc.8'
-  ocp_installer_url: 'https://mirror.openshift.com/pub/openshift-v4/s390x/clients/ocp/'
+  ocp_installer_version: '4.18.8'
+  ocp_installer_base_url: 'https://mirror.openshift.com/pub/openshift-v4'
+  architecture: <multi|s390x>  
   boot_method: <pxe|iso>
 
 # Openshift Settings  
--- a/roles/download_ocp_installer/tasks/main.yaml
+++ b/roles/download_ocp_installer/tasks/main.yaml
@@ -1,43 +1,55 @@
 ---
-- name: Download OpenShift Installer (fips=false).
+- name: Set architecture subfolder for URL (only for multi)
+  set_fact:
+    arch_subdir: "{{ 's390x/' if abi.architecture == 'multi' else '' }}"
+
+- name: Set OCP download variables
+  set_fact:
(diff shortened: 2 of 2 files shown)


inventories/default/group_vars/all.yaml.template
```diff
@@ -229,8 +229,9 @@ day2_compute_node:
 abi:
   flag: False
   ansible_workdir: 'ansible_workdir'
-  ocp_installer_version: '4.15.0-rc.8'
-  ocp_installer_url: 'https://mirror.openshift.com/pub/openshift-v4/s390x/clients/ocp/'
+  ocp_installer_version: '4.18.8'
+  ocp_installer_base_url: 'https://mirror.openshift.com/pub/openshift-v4'
```
v78singh, line 233: is it covered in the doc with the example ?
basically how the user will understand which part of the complete url has to be passed here.

roles/download_ocp_installer/tasks/main.yaml
```diff
@@ -1,43 +1,55 @@
...
+- name: Set architecture subfolder for URL (only for multi)
+  set_fact:
+    arch_subdir: "{{ 's390x/' if abi.architecture == 'multi' else '' }}"
+
+- name: Set OCP download variables
+  set_fact:
+    ocp_download_url: "{{ abi.ocp_installer_base_url }}/{{ abi.architecture }}/clients/ocp/{{ abi.ocp_installer_version }}/{{ arch_subdir }}"
+    ocp_client_tgz: "openshift-client-linux.tar.gz"
+    ocp_installer_tgz: "openshift-install-linux.tar.gz"
+    ocp_install_fips_tgz: "openshift-install-linux"
```
v78singh, line 11: is this value correct ?

Neeraj8418: This code builds the correct OpenShift download URL based on the architecture:

If abi.architecture is "multi", it adds "s390x/" as a subfolder in the URL (since multi-arch URLs include the target arch).

If it's "s390x" or any single-arch, no subfolder is added.

This ensures the playbook works for both multi and s390x by dynamically forming the correct download path.

inventories/default/group_vars/all.yaml.template
Neeraj8418: Sample input i have passed in all.yaml as of now. WIll update the readme aswell. Missed updating that.

roles/download_ocp_installer/tasks/main.yaml
isumitsolanki: Hi @Neeraj8418 , as per our discussion over call. Please do not set variable explicitly. you have to use existing variable and form ocp_installer_url accordingly

PR #404, opened 2025-04-22
Multiarch support for abi

Neeraj8418:
This PR adds support for downloading OpenShift installer and client binaries for both single (s390x) and multi-architecture (multi) configurations.

Changes:
Dynamically constructs the download URL based on the value of abi.architecture.

For multi, appends the architecture-specific subdirectory (s390x/) to the URL.

Ensures the correct binaries are fetched and extracted regardless of the architecture setup.

Purpose:
To streamline support for multi-arch environments while maintaining compatibility with existing single-architecture deployments.


isumitsolanki (changes requested):

isumitsolanki (approved):
Hi @Neeraj8418 , code changes looks good to me now.

Thanks for PR.

PR #406, opened 2025-04-30
Changes to support SNO and 3-NODE OCP Cluster on LPAR

suman-jainkeri:
This PR introduces support for deploying both Single Node OpenShift (SNO) and 3-node OpenShift clusters on IBM Z LPARs.

For these configurations, the all.yaml file includes details for one master node (SNO) or three master nodes (3-node cluster), respectively. In both scenarios, the compute node details are not provided and the tasks related to the creation of compute nodes are intentionally skipped, as they are not required for these cluster types.


AmadeusPodvratnik (approved):
@suman-jainkeri Thank you for the PR. /lgtm

PR #414, opened 2025-06-05
Amrut - Fix for Non s390x install config architecture

amrutp-redhat:
when trying to install x86 ocp cluster install config take s390x. I should be read from all.yaml install config architecture


AmadeusPodvratnik (changes requested):
Hi @amrutp-redhat Thank you for the PR.
Did you check the defaults and there the mechanics how default variables are set (See folder /defaults and /vars). Some of the variables are flatten - see the docs - so if you run into an error because you set env.install_config ... than this is expected and you need to specify install_config.... in your all.yaml.
Pls revalidate your PR. 
In addition you should specify "fix:" as prefix for your PR.
Thank you.

PR #422, opened 2025-07-08
fix: Updated the nameserver of lpar hipersockets agents

jpattara:
Updated the nameserver of lpar hipersockets agents with management cluster nameserver to fix the ingress degrade issue.


veera-damisetti:
Thanks @jpattara  for the PR 
Changes looks good to me 
cc: @AmadeusPodvratnik

PR #424, opened 2025-07-29
fix: Added RoCE interface to the parm file of LPAR while booting

jpattara:
Added the RoCE interface to the parm file of LPAR while booting. 
This will fix the issue of IP address coming for multiple interface while booting the LPAR if we are not mentioning the specific interface in the parm file.


veera-damisetti (changes requested):
@jpattara 
Thanks for the PR 
Can you please verify the conditions once . 
we need to remove {% endif %} also as we are  removing if condition there

veera-damisetti:
Thanks @jpattara for changes 
looks good to me

veera-damisetti:
cc: @sandisamp

PR #467, opened 2026-03-23
fix: include cluster metadata name in bastion lookup

Neeraj8418:
Description
During the OpenShift Agent-Based Installation (ABI) on s390x, the check_dns role was failing to validate the Bastion host's IP address. This occurred because the dig command was looking for the Bastion at {{ hostname }}.{{ base_domain }}, whereas the internal DNS zone created for the cluster expects the format {{ hostname }}.{{ cluster_metadata_name }}.{{ base_domain }}.

This PR updates the lookup logic to include the cluster metadata name, ensuring that internal connectivity checks pass before the LPARs attempt to pull the MachineOS images.

Changes
roles/check_dns/tasks/main.yaml

Logic: Updated shell tasks to include {{ env.cluster.networking.metadata_name }} in the FQDN string for both standard and Internal_IP lookup scenarios.

Impact
Prevents failed_when triggers during the infrastructure validation phase.

Ensures consistent DNS resolution across the bastion and the s390x control/compute nodes.

Testing Performed
Verified dig output manually on the abi-bastion host.

Confirmed the bastion_lookup.stdout matches the expected env.bastion.networking.ip.

### DIFF
--- a/roles/check_dns/tasks/main.yaml
+++ b/roles/check_dns/tasks/main.yaml
@@ -1,21 +1,24 @@
 ---
-
 - name: Check internal cluster DNS resolution for the bastion
   tags: check_dns, dns
-  shell: "dig +short {{ env.bastion.networking.hostname }}.{{ env.bastion.networking.base_domain }} | tail -n1"
+  ansible.builtin.shell: |
+    set -o pipefail
+    dig +short {{ env.bastion.networking.hostname }}.{{ env.cluster.networking.metadata_name }}.{{ env.bastion.networking.base_domain }} | tail -n1
   register: bastion_lookup
   failed_when: env.bastion.networking.ip != bastion_lookup.stdout
-  when: env.bastion.networking.internal_ip is not defined or env.bastion.networking.internal_ip == None
+  when: env.bastion.networking.internal_ip is not defined or env.bastion.networking.internal_ip == none
 
 - name: Check internal cluster DNS resolution for the bastion with Internal_IP
   tags: check_dns, dns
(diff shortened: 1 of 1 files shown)


roles/check_dns/tasks/main.yaml
```diff
@@ -1,21 +1,24 @@
 ---
-
 - name: Check internal cluster DNS resolution for the bastion
   tags: check_dns, dns
-  shell: "dig +short {{ env.bastion.networking.hostname }}.{{ env.bastion.networking.base_domain }} | tail -n1"
+  ansible.builtin.shell: |
+    set -o pipefail
+    dig +short {{ env.bastion.networking.hostname }}.{{ env.cluster.networking.metadata_name }}.{{ env.bastion.networking.base_domain }} | tail -n1
```
veera-damisetti, line 6: Hi @Neeraj8418 
why did we give here 
`{{ env.bastion.networking.hostname }}.{{ env.cluster.networking.metadata_name }}.{{ env.bastion.networking.base_domain }}`

Are we adding dns resolution to hostname.metadataname.basedomain ?

Monisha0079: yes, we are using `hostname.metadata_name.base_domain` because thats how the dns record is actually created (ex:agent-bastion.agent.example.com) so the validation needs to check the same format. If we only use `hostname.base_domain` it doesn’t match the actual DNS record and the check fails
@Neeraj8418 and I tested this using dig, and it worked correctly with the updated format.

veera-damisetti: Sure @Monisha0079 
Thanks

veera-damisetti (approved):

PR #472, opened 2026-03-27
fix: issue_471; wrong var name for subscription

AmadeusPodvratnik:
ifix for wrong variable name. 
Regression: yes

AI involved: No


Sravikaz (approved):
looks good

PR #474, opened 2026-03-27
fix: issue 473; issue with range command.

AmadeusPodvratnik:
This fixes the issue with range command with respect to container, python and ansible plays.

AI involved: No


abdulkandathil (approved):
/lgtm

Already tested and confirmed.

PR #475, opened 2026-03-30
feat: Enable rootDeviceHints support for ABI

Neeraj8418:
Enable support for rootDeviceHints param for lpar boot using liveDisk.
This will also be used to add support for multipath for ABI. (zVM & LPAR)

### DIFF
--- a/inventories/default/group_vars/all.yaml.template
+++ b/inventories/default/group_vars/all.yaml.template
@@ -142,7 +142,8 @@ env:
 #         - X
 #         - X
 #         - X
-
+        root_device_hints:
+          - ''
 # Section 9 - Compute Nodes
       compute:
         disk_size: 120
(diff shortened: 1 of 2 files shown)


inventories/default/group_vars/all.yaml.template
```diff
@@ -142,7 +142,8 @@ env:
 #         - X
 #         - X
 #         - X
-
+        root_device_hints:
```
AmadeusPodvratnik, line 145: Is this something valid for all nodes? If yes, than every node de need to contain this hint. 
Doc missing. You need to describe this parameter, too.

AmadeusPodvratnik (changes requested):
@Neeraj8418 Thank you for the PR.
Pls see my comment.

Neeraj8418:
Not needed anymore

PR #486, opened 2026-04-22
fix: Add setting for delay for check nodes

AmadeusPodvratnik:
Environments having performance issue are delayed rolling out the OCP cluster and AOP could fail because of timeouts. This PR makes the wait time for checking the nodes configurable and set the default to 30s instead of 20s.

AI involved: No


abdulkandathil (approved):
/lgtm

Issue #498, opened 2026-05-20
feat: Create local registry and us it for UPI install.


AmadeusPodvratnik:
One of the disconnected use cases is to build a local registry and pull all images to be able to install OCP disconnected.

The new playbooks will replace the old playbooks but will be removed later.

dk5488:
@AmadeusPodvratnik Hi can I work on this next ?

AmadeusPodvratnik:
@dk5488 
There is already some development done. You can find it here:
https://github.com/IBM/Ansible-OpenShift-Provisioning/pull/499
This is a very broad topic and should replace the existing disconnected install at the end. 
I you want you can take it and extend it (get the code and fix the merge issues and / or add it to the latest codebae). In addition the local registry should also being configured as a path-through registry.
But a possible PR requires "some" testing effort you need to provide.
Pls have a look at this PR / branch. I postponed it due to some priority shifts.

PR #499, opened 2026-05-20
fix: Create a local registry on the bastion host to install OCP cluster airgapped. 

AmadeusPodvratnik:
This is the implementation to create a local registry and install OCP from there.
This new playbook will replace the old playbooks. In addition the flag for disconnected.enabled is being pushed to all.yaml for better control (centralized).

Fix for issue: https://github.com/IBM/Ansible-OpenShift-Provisioning/issues/498

This PR also contains some fixes and variables checks causing failures  (not only in the context of disconnected).

Tested against 3 HA cluster (KVM) and 5 node cluster using NAT on a single KVM LPAR.

AI involved: Yes


AmadeusPodvratnik:
Line to long checks will be fixed later ... will concentrate on the fqcn names

AmadeusPodvratnik:
There are to much lint errors to be fixed at once ... will do it in a second round.

PR #503, opened 2026-06-03
feat: Implementation of HCP with Kubevirt support

jpattara:
Hosted Control Planes (HCP) with KubeVirt on IBM Z enables OpenShift control planes to run as KubeVirt virtual machines on IBM Z (s390x), providing efficient, scalable, and infrastructure-isolated cluster management.

As part of this PR, below steps are taken care of:

- Provisioning of HPP storage class
- Creation of HCP cluster with type KubeVirt
- Deletion of HCP cluster
- Optional MCE, MetalLB & LSO custom catalog source support
- FIPS validation support
- Complete documentation for HCP KubeVirt workflow


inventories/default/group_vars/hcp-kubevirt.yaml.template
```diff
@@ -41,25 +43,37 @@ hcp:
         - #disk1
         - #disk2
         - #disk3
+  catalogsource_image: ""  # Required only if we need to create the Catalogsource for installing lso operator
```
isumitsolanki: @jpattara , Why we are initialing `catalogsource_image` same parameter at two place ?

inventories/default/group_vars/hcp-kubevirt.yaml.template
```diff
@@ -22,6 +23,7 @@ hcp:
     ip_pool: #minimum 2 free IPs are required in the subnet for MetalLB to work properly. You can provide a single IP or IP ranges. For example:
     #- x.x.x.x-x.x.x.x
     #- x.x.x.x-x.x.x.y
+  catalogsource_image: ""  # Required only if we need to create the Catalogsource for installing metallb operator
```
isumitsolanki: .

playbooks/hcp_kubevirt_prereqs_setup.yaml
```diff
@@ -1,13 +1,4 @@
 ---
-# Setup inventory and passwordless SSH for bastion
-- name: Setup inventory and passwordless SSH
-  hosts: localhost
-  vars_files:
-    - "{{ inventory_dir }}/group_vars/secrets.yaml"
-    - "{{ inventory_dir }}/group_vars/hcp-kubevirt.yaml"
-  roles:
-    - create_inventory_setup_hcp_kubevirt
```
isumitsolanki, line 9: Why do we have removed `create_inventory_setup_hcp_kubevirt` role ?? this role is responsible for creating inventory file for hcp agent as well.

isumitsolanki (commented):
Hi @jpattara , Thanks for PR please check my comments

inventories/default/group_vars/hcp-kubevirt.yaml.template
jpattara: @isumitsolanki Thanks for the review.

`catalogsource_image` is intended to be defined under both the MetalLB and LSO operator sections, as each operator may require a custom CatalogSource image when installing on candidate versions of OCP.

We have observed that the MetalLB and LSO operators are not always available in the default CatalogSource for candidate builds, whereas they are available in stable releases. To support operator installation in candidate versions, a custom `catalogsource_image` can be specified for each operator independently.

The current duplication is due to an indentation issue in the template, which makes it appear as though the same parameter is defined at the same level. The parameter should actually reside within the respective `metallb` and `lso` sections. I'll correct the indentation in the next update.

playbooks/hcp_kubevirt_prereqs_setup.yaml
jpattara: @isumitsolanki I have added the inventory setup role as a separate playbook in the main `hcpvirt.yaml` workflow. This provides the flexibility to execute the inventory and SSH key setup independently when needed, while still allowing it to be run as part of the complete end-to-end deployment flow.

```
- name: Setup inventory and SSH keys
  import_playbook: setup_inventory_hcp_kubevirt.yaml

- name: Run HCP KubeVirt prerequisites setup
  import_playbook: hcp_kubevirt_prereqs_setup.yaml

- name: Create HCP KubeVirt cluster
  import_playbook: create_hcpvirt.yaml
```

isumitsolanki: Thanks @jpattara for addressing all the comments. 
Can you git squash in single commit.

PR #508, opened 2026-06-25
fix: Agent-Based Installer robustness fixes (workdir, DNS, ISO boot, ssh_agent)

vinayakray19:
## Summary

A set of small, independent robustness fixes to the Agent-Based Installer (ABI)
flow, found while deploying OpenShift 4.22 ABI clusters on z/LinuxONE KVM. Each
fix is a separate commit and none change default behaviour for existing users.

### 1. Configurable OCP install work directory (`refactor`)
- Adds `roles/common/tasks/resolve_ocp_install_workdir.yaml` which resolves a
  single `ocp_install_workdir` fact (defaulting to `abi.ansible_workdir`) and
  asserts it is set.
- `prepare_configs`, `setup_params`, `create_agent`, and `abi_install_complete`
  now use `ocp_install_workdir` instead of hardcoding `~/ansible_workdir` or
  repeating `abi.ansible_workdir`. This also fixes `prepare_configs`, which
  previously created/deleted a hardcoded `~/ansible_workdir` that did not match
  the configured `abi.ansible_workdir`.

### 2. Make bastion `nameserver2` optional in DNS config (`fix`)
- `roles/dns/templates/dns-named.conf.j2` referenced
  `env.bastion.networking.nameserver2` unconditionally, causing a Jinja
  `UndefinedError` when only one upstream nameserver is configured. The second
  forwarder is now guarded and falls back to the cluster forwarder.

### 3. Keep agent VMs running after ISO boot (`fix`)
- The ISO `virt-install` tasks used `--wait=-1`, which left the ISO-booted agent
  VMs powered off and stalled the install. `--wait=-1` is removed from both the
  control and compute ISO commands, and two tasks were added: one waits until
  every VM definition is registered, and one ensures all agent VMs are running.

### 4. Run `ssh_agent` play as the login user with secrets (`fix`)
- `playbooks/create_abi_cluster.yaml`: the "Setup SSH agent" play now sets
  `become: false` and loads `group_vars/all.yaml` and `group_vars/secrets.yaml`,
  so the `ssh_agent` role runs as the bastion login user and can see the
  secret-backed variables it needs.

## Test plan
- [x] Deployed two OpenShift 4.22.0-ec.3 ABI clusters end-to-end on z/LinuxONE
      KVM (ISO boot) with these changes.
- [ ] Confirm PXE boot path is unaffected (only ISO `virt-install` lines changed).
- [ ] Confirm existing inventories with `nameserver2` set still render identical
      `named.conf` forwarders.

### DIFF
--- a/roles/boot_abi_agents/tasks/main.yml
+++ b/roles/boot_abi_agents/tasks/main.yml
@@ -79,8 +79,7 @@
     --disk size={{ env.cluster.nodes.control.disk_size }} \
     --network network={{ env.vnet_name }}{{ (',mac=' + env.cluster.nodes.control.mac[i]) }} \
     --graphics none \
-    --noautoconsole \
-    --wait=-1 \
+    --noautoconsole
   async: 3600
   poll: 0
   with_sequence: start=0 end={{ (env.cluster.nodes.control.hostname | length) - 1 }} stride=1
@@ -149,8 +148,7 @@
     --disk size={{ env.cluster.nodes.compute.disk_size }} \
     --network network={{ env.vnet_name }}{{ (',mac=' + env.cluster.nodes.compute.mac[i]) }} \
     --graphics none \
-    --noautoconsole \
-    --wait=-1 \
+    --noautoconsole
   async: 3600
   poll: 0
   with_sequence: start=0 end={{ (env.cluster.nodes.compute.hostname | length) - 1 }} stride=1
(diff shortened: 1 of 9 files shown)


roles/boot_abi_agents/tasks/main.yml
```diff
@@ -149,8 +148,7 @@
     --disk size={{ env.cluster.nodes.compute.disk_size }} \
     --network network={{ env.vnet_name }}{{ (',mac=' + env.cluster.nodes.compute.mac[i]) }} \
     --graphics none \
-    --noautoconsole \
-    --wait=-1 \
+    --noautoconsole
```
v78singh: Why --wait=-1 is removed , this can break ABI flow

vinayakray19: Good point — I've made this configurable instead of removing it outright (0c6267b).

By default `abi.iso_virt_install_wait: true`, so `--wait=-1` is still passed for the ISO `virt-install` commands and the existing ABI flow is unchanged.

Why the toggle: these ISO tasks run with `async: 3600, poll: 0` (fire-and-forget). On slower hosts the agent install runs past that async window, and when Ansible reaps the async job at the 3600s timeout it tears down the still-attached `virt-install` process (kept alive by `--wait=-1`), which powers the agent VM off mid-install. Setting `abi.iso_virt_install_wait: false` lets operators avoid that on affected hosts.

I also kept the new "Ensure ABI agent VMs are running after ISO boot" task as a safety net so the VMs are guaranteed to be up before the install proceeds, regardless of the toggle.

Neeraj8418:
This PR is not needed.

sandisamp:
@vinayakray19 please close this PR as indicated by @Neeraj8418

vinayakray19:
Not needed as confirmed by maintainers.

PR #526, opened 2026-07-09
fix: day-2 compute node additions for multi-arch support

### LINKED ISSUE #524: fix: day-2 compute node playbook broken for multi-arch (aarch64/x86_64/s390x) clusters
## Problem

The `playbooks/create_compute_node.yaml` playbook and its supporting roles have couple of bugs
that prevent successful day-2 compute node addition, particularly on multi-arch clusters
(e.g. aarch64 KVM hosts alongside s390x nodes).

### 1. `roles/create_compute_node/tasks/main.yaml`

- `/var/www/html/bin` directory not created before `get_url` download — task fails if directory is absent
- Nameserver kernel arg uses comma separator (`nameserver=x,y`) instead of correct dracut syntax (`nameserver=x nameserver=y`)
- Disk config always applied s390x-style options (`cache=none,io=native`) regardless of architecture — breaks aarch64/x86_64
- Network config always used s390x vnet instead of macvtap direct bridge for other architectures
- `install_dev` variable (`param_compute_node.install_dev | default('vda')`) was hardcoded to `vda`, silently ignoring the node config value
- `virt-install` timeout was 360s — too short for aarch64 installs, causing spurious failures

### 2. `roles/dns_update/tasks/add.yaml`

- Forward DNS zone file path used `.db` extension instead of `.<base_domain>.zone`
- Reverse DNS zone file path used `.rev` instead of the correct `in-addr.arpa.zone` with reversed IP octets
- PTR record Jinja2 syntax used `.3` (invalid attribute access on a list) instead of `[3]` (bracket index)

### 3. `inventories/default/group_vars/all.yaml.template`

- `disk_pool` field missing from the `day2_compute_node` section — users had no template reference for this field, which is required on non-s390x architectures

## Impact

- Day-2 compute node additions fail or produce nodes that never reach `Ready` state
- Multi-arch clusters (aarch64 KVM hosts) broken by wrong `virt-install` disk/network arguments
- DNS entries for new nodes written to wrong zone files, causing name resolution failures for the new node

## Fix

See PR #526  — branch `sravika-dev`

Sravikaz:
Fixes #524

## Summary

Fixes multiple bugs in the day-2 compute node playbook that prevented successful
node addition on multi-arch clusters (aarch64 KVM hosts with s390x hosts).

### `playbooks/create_compute_node.yaml`
- Fix example command to use correct playbook name and `--ask-vault-pass`
- Fix error message to reference correct playbook name instead of old `add_compute_node.yaml`

### `roles/create_compute_node/tasks/main.yaml`
- Add task to ensure `/var/www/html/bin` directory exists before RHCOS image download
- Fix nameserver kernel arg separator: `,` → `nameserver=` (correct dracut/kernel cmdline syntax)
- Add arch-aware disk config:
  - `s390x`: `cache=none,io=native` with vdisk pool
  - `aarch64/x86_64`: `discard=unmap,bus=virtio` with default pool
- Add arch-aware network config:
  - `s390x`: `--network network={{ vnet_name }}`
  - `aarch64/x86_64`: `--network type=direct,source={{ macvtap_dev }},source_mode=bridge`
- Restore `install_dev` variable: `param_compute_node.install_dev | default('vda')` (was hardcoded to `vda`)
- Increase `virt-install` timeout: `360s` → `1800s` (prevents spurious failures on aarch64)

### `roles/dns_update/tasks/add.yaml`
- Fix forward DNS zone file path: `.db` → `.<metadata_name>.<base_domain>.zone`
- Fix reverse DNS zone file path: `.rev` → `<reversed-octets>.in-addr.arpa.zone`
- Fix PTR record Jinja2 syntax: `.3` → `[3]` (bracket index — attribute access on a list is invalid)

### `inventories/default/group_vars/all.yaml.template`
- Add `disk_pool` field to `day2_compute_node` section so users have a template reference

## Testing

Validated on a multi-arch cluster:
- aarch64 KVM host 
- s390x control and compute nodes
- `worker-3` successfully joined the cluster and reached `Ready` state

### DIFF
--- a/inventories/default/group_vars/all.yaml.template
+++ b/inventories/default/group_vars/all.yaml.template
@@ -228,6 +228,7 @@ day2_compute_node:
   hostname:
   host_user:
   host_arch:
+  disk_pool:
   install_dev: # (Optional) Installation device, defaults to 'vda' if not specified
 
 
--- a/docs/set-variables-group-vars.md
+++ b/docs/set-variables-group-vars.md
@@ -201,7 +201,8 @@
 **day2_compute_node.hostname** | The hostname of the KVM host | kvm-host-01
 **day2_compute_node.host_user** | KVM host user which is used to create the VM | root
 **day2_compute_node.host_arch** | KVM host architecture.  | s390x
-**day2_compute_node.install_dev** | <b>(Optional)</b> Installation device for CoreOS installation. Defaults to 'vda' if not specified. | vda
+**day2_compute_node.disk_pool** | <b>(Optional)</b> Name of the libvirt storage pool to use for the compute node VM disk. On s390x defaults to `<metadata_name>-vdisk`. On all other architectures (aarch64, x86_64) defaults to `default`. Only set this if your KVM host uses a custom pool name. | default
+**day2_compute_node.install_dev** | <b>(Optional)</b> Installation device for CoreOS installation. Defaults to 'vda' if not specified. Leave blank or omit to use the default. | vda
 
 ## 14 - (Optional) Agent Based Installer
 **Variable Name** | **Description** | **Example**
--- a/playbooks/create_compute_node.yaml
+++ b/playbooks/create_compute_node.yaml
@@ -16,7 +16,7 @@
 #
 # Execute the playbook with '--extra-vars' option.
 # E.g.:
-# ansible-playbook playbooks/add_compute_node.yaml --extra-vars "@extra-cnode1.yml"
+# ansible-playbook playbooks/create_compute_node.yaml --ask-vault-pass
 
 - name: Add an additional compute node
   # Select bastion host
@@ -33,7 +33,7 @@
           ansible.builtin.debug:
             msg:
               - "ERROR: Variable 'day2_compute_node' is not defined!"
-              - "Execute: 'ansible-playbook playbooks/add_compute_node.yaml --extra-vars \"@extra-cnode.yml\"'"
+              - "Execute: 'ansible-playbook playbooks/create_compute_node.yaml'"
         - name: Abort playbook
           ansible.builtin.fail:
             msg: "See above error!"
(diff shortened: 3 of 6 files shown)


AmadeusPodvratnik (changes requested):
Thx @Sravikaz for the PR. Could you pls add the documentation of this new parameter.
Didn't you mentioned about arm architecture and the missing of macvtap support? I mean it would be good to mention this in the documentation, too.
Can you pls recheck the named stuff because the current entries for named are addressed like this:
//forward zone
zone "a311-odf.lnxero1.boe" IN {
     type master;
     file "/var/named/a311-odf.db";
     allow-update { any; };
     allow-query { any; };
};

The code in dns_update should add a line after:
;entries for the compute nodes

Sravikaz:
I have updated the named stuff accordingly and I've also reverted the changes with respect to macvatp as the arm kvm host has the macvtap libvirt network created now.

inventories/default/group_vars/all.yaml.template
```diff
@@ -228,6 +228,7 @@ day2_compute_node:
   hostname:
   host_user:
   host_arch:
+  disk_pool:
```
AmadeusPodvratnik, line 231: Can you pls add the documentation for this field to the set-variables-group-vars.yaml.

AmadeusPodvratnik (changes requested):

AmadeusPodvratnik:
> I have updated the named stuff accordingly and I've also reverted the changes with respect to macvatp as the arm kvm host has the macvtap libvirt network created now.

So I assume this is an manual step as before, correct?

inventories/default/group_vars/all.yaml.template
Sravikaz: done

Sravikaz:
> > I have updated the named stuff accordingly and I've also reverted the changes with respect to macvatp as the arm kvm host has the macvtap libvirt network created now.
> 
> So I assume this is an manual step as before, correct?

there is no manual change needed to the nameserver filename if you have installed OCP with AOP.
As I have installed my cluster with zVM_Automation, the nameserver file names were different, hence I had to manually update those file names to what AOP is using.

AmadeusPodvratnik (approved):
@Sravikaz Thank you for the update. /lgtm

PR #540, opened 2026-08-04
feat: fix: NAT network support - suppress host key checking and pre-authorised key

Neeraj8418:
controller key in bastion kickstart

- ansible.cfg: add StrictHostKeyChecking=no and UserKnownHostsFile=/dev/null to ssh_args so stale known_hosts entries never block playbook reconnects after bastion is recreated (common in NAT/KVM setups)

- rhel9-bastion-ks.cfg.j2: inject the Ansible controller public key (path_to_key_pair) into /root/.ssh/authorized_keys during kickstart %post so Ansible can connect via ProxyJump on first contact without a separate ssh-copy-id step

Fixes recurring 'REMOTE HOST IDENTIFICATION HAS CHANGED' and 'Permission denied (publickey)' failures when running ABI cluster installation over NAT network mode.

### DIFF
--- a/roles/create_bastion/templates/rhel9-bastion-ks.cfg.j2
+++ b/roles/create_bastion/templates/rhel9-bastion-ks.cfg.j2
@@ -101,6 +101,10 @@ pkg-config
 # Allow root login, required change for RHEL9
 sed -i -e '/PermitRootLogin/ c\PermitRootLogin yes' /etc/ssh/sshd_config
 
+# Pre-authorize the Ansible controller's SSH public key so Ansible can connect directly via ProxyJump
+echo "{{ lookup('file', path_to_key_pair) }}" >> /root/.ssh/authorized_keys
+chmod 600 /root/.ssh/authorized_keys
+
 # Basic /root/.ssh/config setup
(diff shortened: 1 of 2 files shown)


roles/create_bastion/templates/rhel9-bastion-ks.cfg.j2
```diff
@@ -101,6 +101,12 @@ pkg-config
 # Allow root login, required change for RHEL9
 sed -i -e '/PermitRootLogin/ c\PermitRootLogin yes' /etc/ssh/sshd_config
 
+# Pre-authorize the Ansible controller's SSH public key so Ansible can connect directly via ProxyJump
+mkdir -p /root/.ssh
```
veera-damisetti, line 105: Do we need mkdir statement here ?

Neeraj8418: removed and addressed

veera-damisetti (approved):

veera-damisetti:
Thanks @Neeraj8418  for the PR 
cc: @sandisamp

PR #543, opened 2026-08-12
fix: add timeout parameter for create nodes.

AmadeusPodvratnik:
On slower environments and running in pipelines creating the nodes could happen much slower than expected and the timeout of 360s is not enough.
This PR introduces a timeout parameter having the default value of 360. 
This PR was tested against a single KVM LPAR successfully. Network configuration was NAT.


AmadeusPodvratnik:
The lint errors (lines to long) will be fixed in a later PR.

Issue #546, opened 2026-08-26
bug: on slow environments the bootstrap process could take longer and scripts are failing with timeout error.


AmadeusPodvratnik:
On slower environments the whole bootstrap and creation process could take much longer than expected and the scripts are failing with timeout error.

AmadeusPodvratnik:
Fixed with https://github.com/IBM/Ansible-OpenShift-Provisioning/pull/547

PR #547, opened 2026-08-27
fix: Add more wait time for slower envs.

AmadeusPodvratnik:
Fix for https://github.com/IBM/Ansible-OpenShift-Provisioning/issues/546
The wait time for bootstrapping need to be extended, too. On slower environments the scripts might fail because the overall processing time is slow and the operators do not progressing in a timely manner.
Make the wait time for creating the bootstrap configurable (the same parameter used like the other nodes) and prolong the wait time for the bootstrap process in general.
Tested with two different KVM LPARs (NAT).
 
AI involved: no


AmadeusPodvratnik:
@Sravikaz If possible pls do the review on this PR. Thank you.

AmadeusPodvratnik:
The line to long lint errors will be fixed in one of upcoming PRs.

Sravikaz (approved):
LGTM

PR #549, opened 2026-08-27
fix: Set force flag for get rhcos image.

AmadeusPodvratnik:
In case of changing OCP versions to be installed, the RHCOS image need to be installed again in 5_setup_bastion playbook. To achive this, the force flag is being set to true.
AI involved: No


AmadeusPodvratnik:
The missing lint fixes will be done in a later PR.

Sravikaz (approved):
LGTM

PR #550, opened 2026-09-18
feat: Install tigervnc for better administration (#500)

### LINKED ISSUE #500: feat: Install tigervnc for better administration.
To administrate the KVM much better an option is to install and configure / enable tigervnc.

At the moment no UI support to administrate a e.g. KVM Server is automatically configured. 
Due to the fact that creating a KVM LPAR is possible, some sort of administration support should be present. One option is to install and configure tigerVNC.

dk5488:
## Description
Resolves #500

To improve KVM administration, this PR automates the installation, configuration, and enablement of **TigerVNC Server** on KVM hypervisors during execution of `playbooks/3_setup_kvm_host.yaml`.

Currently, KVM hosts only have CLI access or the Cockpit web socket enabled. With TigerVNC configured, administrators can easily connect with a graphical VNC client (e.g. TigerVNC Viewer) at `<kvm-host-ip>:5901` (display `:1`) to manage virtual machines and use graphical utilities like `virt-manager`.

### Key Changes
1. **Package Defaults**:
   - Added `tigervnc-server` to `pkgs_kvm` in `roles/install_packages/defaults/main.yaml`.
2. **New Role (`roles/tigervnc`)**:
   - `defaults/main.yaml`: Provides sensible defaults (`tigervnc_display: 1`, `port: 5901`, `geometry: 1920x1080`, `session: gnome`) and smart password fallback.
   - `templates/vnc_config.j2`: Generates `~/.vnc/config` with session and resolution settings.
   - `tasks/main.yaml`:
     - Ensures `tigervnc-server` is installed.
     - Configures user mapping in `/etc/tigervnc/vncserver.users` (`:1=<user>`).
     - Prepares user's `~/.vnc/` directory with secure permissions (`0700`).
     - Sets the VNC password using `vncpasswd -f` with `no_log: true` and `0600` permissions.
     - Opens TCP port `5901` in `firewalld`.
     - Enables and starts `vncserver@:1.service` via `systemd`.
3. **Playbook Integration**:
   - Integrated `- role: tigervnc` into `playbooks/3_setup_kvm_host.yaml`.
4. **Kickstart Configuration**:
   - Added port `5901` to the firewall rule in `roles/update_cfgs/templates/kvm_host.cfg.j2` for newly provisioned KVM hosts.
5. **Configuration Templates**:
   - Added optional `tigervnc` block under `env.z` in `inventories/default/group_vars/all.yaml.template`.
   - Added optional `vault_vnc_pass` in `inventories/default/group_vars/secrets.yaml.template`.
6. **Documentation**:
   - Updated `docs/set-variables-group-vars.md` with new parameters and updated `pkgs_kvm` table.
   - Updated `docs/run-the-playbooks.md` to document TigerVNC server outcome under Step 3.

### Testing & Verification
- Validated all modified and newly created YAML files with Python YAML syntax check (`yaml.safe_load`).
- Verified compliance with repository `.ansible-lint` conventions and formatting standards.


docs/set-variables-group-vars.md
```diff
@@ -245,7 +250,7 @@
 :--- | :--- | :---
 **pkgs_galaxy** | A list of Ansible Galaxy collections that will be installed during the setup playbook. The collections listed are required. | [ ibm.ibm_zhmc, community.general, community.crypto, ansible.posix, community.libvirt ]
 **pkgs_controller** | A list of packages that will be installed on the machine running Ansible during the setup playbook. | [ openssh, expect, sshuttle ]
-**pkgs_kvm** | A list of packages that will be installed on the KVM Host during the setup_kvm_host playbook. | [ libguestfs, libvirt-client, libvirt-daemon-config-network, libvirt-daemon-kvm, cockpit-machines, libvirt-devel, virt-top, qemu-kvm, python3-lxml, cockpit, lvm2 ]
```
AmadeusPodvratnik: You are removing libvirt-devel, why?

docs/set-variables-group-vars.md
```diff
@@ -245,7 +250,7 @@
...
+**pkgs_kvm** | A list of packages that will be installed on the KVM Host during the setup_kvm_host playbook. | [ libguestfs, libvirt-client, libvirt-daemon-config-network, libvirt-daemon-kvm, cockpit-machines, virt-top, qemu-kvm, python3-lxml, cockpit, lvm2, httpd, tigervnc-server ]
```
AmadeusPodvratnik: to enable tigervnc and use it, there is GNOME needed. I think there is following needed (not sure if both Graphical groupinstall are needed):
dnf groupinstall "Graphical Administration Tools"
dnf groupinstall graphical-server-environment
dnf groupinstall GNOME

It would be good to ensure the GNOME install.

Not sure if the regular ansible module could handle group installs like the dnf modules can do. E.g.:
- name: Install the 'Graphical Administration Tools' package group
  ansible.builtin.dnf:
    name: '@Graphical Administration Tools'
    state: present

inventories/default/group_vars/all.yaml.template
```diff
@@ -31,6 +31,13 @@ env:
 #      ip:
 #      user:
 #      pass: "{{ vault_lpar3_pass }}"
+    # (Optional) TigerVNC Remote Desktop Administration for KVM host
+    tigervnc:
+      enabled: True
+      display: 1
+      geometry: 1920x1080
+      session: gnome
+      # pass: "{{ vault_vnc_pass }}"
```
AmadeusPodvratnik, line 40: There is no readonly pwd set. Make it sense to add a readonly, too?

inventories/default/group_vars/secrets.yaml.template
```diff
@@ -25,6 +25,9 @@ vault_lpar2_root_pass: #X
 vault_lpar3_user_pass: #X
 vault_lpar3_root_pass: #X
 
+# Section 2.3 - TigerVNC Password (optional, defaults to LPAR user password if not set)
+vault_vnc_pass: #X
```
AmadeusPodvratnik, line 29: See the comment for all.yaml.template

playbooks/3_setup_kvm_host.yaml
```diff
@@ -85,6 +85,8 @@
       when: installation_type | lower in ['kvm', 'lpar']
     - role: httpd
       when: installation_type | lower in ['kvm', 'lpar']
+    - role: tigervnc
+      when: installation_type | lower in ['kvm', 'lpar']
```
AmadeusPodvratnik: tigervnc is only valid for kvm. The when statement for lpar is wrong and a bug. 
lpar installation type will not do a setup_kvm because the entire lpar will be used as a node.
I will open a bug for this. So just remove the lpar in your when clause.

roles/install_packages/defaults/main.yaml
```diff
@@ -1,4 +1,4 @@
 # Packages
 pkgs_controller: [ openssh, expect, sshuttle ]
-pkgs_kvm: [ libguestfs, libvirt-client, libvirt-daemon-config-network, libvirt-daemon-kvm, cockpit-machines, virt-top, qemu-kvm, python3-lxml, cockpit, lvm2, httpd ]
+pkgs_kvm: [ libguestfs, libvirt-client, libvirt-daemon-config-network, libvirt-daemon-kvm, cockpit-machines, virt-top, qemu-kvm, python3-lxml, cockpit, lvm2, httpd, tigervnc-server ]
```
AmadeusPodvratnik: Here the same packages for tigervnc (incl. gnome) should be done.

roles/update_cfgs/templates/kvm_host.cfg.j2
```diff
@@ -32,7 +32,7 @@ lang {{ env.language }}
 # Network information (will fill in during update_cfgs role)
 
 # Firewall and SELinux
-firewall --enabled --http --ftp --smtp --ssh --port=443,9090,123
+firewall --enabled --http --ftp --smtp --ssh --port=443,9090,123,5901
```
AmadeusPodvratnik: you have the 5901 value hardcoded but you need to take what the user configures in all.yaml (env.z.tigervnc.display) - but only if tigervnc should be installed.

roles/tigervnc/tasks/main.yaml
```diff
@@ -0,0 +1,85 @@
...
+    echo "{{ tigervnc_password }}" | vncpasswd -f > {{ tigervnc_user_home.stdout }}/.vnc/passwd
+    chmod 0600 {{ tigervnc_user_home.stdout }}/.vnc/passwd
+    chown {{ tigervnc_user }}:{{ tigervnc_user }} {{ tigervnc_user_home.stdout }}/.vnc/passwd
+  no_log: true
+  when:
+    - tigervnc_enabled | bool
+    - tigervnc_password is defined
+    - tigervnc_password | length > 0
+
+- name: Open VNC port in firewall
```
AmadeusPodvratnik, line 83: There is a template for the firewall opens a the tigervnc port. Is this needed here?

AmadeusPodvratnik (changes requested):
Hi @dk5488 , thank you for this PR. It looks very good but I have some minor comments.

docs/set-variables-group-vars.md
dk5488: Good point! On RHEL 8/9, `"@Server with GUI"` provides the full GNOME desktop environment and `"@Graphical Administration Tools"` installs graphical tools like `virt-manager`.

Because `roles/install_packages` uses `ansible.builtin.package` with `state: latest` in a loop (which doesn't reliably support DNF environment/package groups starting with `@`), and to avoid forcing headless KVM hosts to download ~1.5 GB of desktop packages when TigerVNC is disabled, I have added a dedicated task in `roles/tigervnc/tasks/main.yaml` using `ansible.builtin.dnf` with `state: present`.

It installs `["@Server with GUI", "@Graphical Administration Tools"]` conditionally when `tigervnc_enabled: true` and `tigervnc_session == 'gnome'`.

inventories/default/group_vars/all.yaml.template
dk5488: Makes complete sense. I've added `# readonly_pass: "{{ vault_vnc_readonly_pass }}"` under the `tigervnc` section in `all.yaml.template`. 

The `roles/tigervnc` defaults and password tasks have been updated so that when this variable is provided, `vncpasswd -f` receives both the full-access password and the secondary view-only password via stdin.

inventories/default/group_vars/secrets.yaml.template
dk5488: Added `vault_vnc_readonly_pass: #X` under Section 2.3 in `secrets.yaml.template`.

playbooks/3_setup_kvm_host.yaml
dk5488: Understood and updated! Changed the condition to:
`when: installation_type | lower == 'kvm'`

roles/install_packages/defaults/main.yaml
dk5488: As mentioned in the documentation comment, installing GNOME package groups (`@Server with GUI` and `@Graphical Administration Tools`) is now handled inside `roles/tigervnc/tasks/main.yaml` using `ansible.builtin.dnf` with `state: present`.

This keeps `pkgs_kvm` lightweight and ensures that if a user sets `tigervnc: enabled: False`, their KVM host remains a minimal server without installing unwanted GUI packages.

roles/update_cfgs/templates/kvm_host.cfg.j2
dk5488: Updated! Replaced the hardcoded `5901` with dynamic Jinja2 templating:

`firewall --enabled --http --ftp --smtp --ssh --port=443,9090,123{% if env.z is defined and env.z.tigervnc is defined and (env.z.tigervnc.enabled is not defined or env.z.tigervnc.enabled in [True, 'True', 'true', 1]) %},{{ 5900 + (env.z.tigervnc.display | default(1) | int) }}{% endif %}`

This dynamically computes `5900 + display` and only appends the port if `tigervnc` is defined and enabled.

roles/tigervnc/tasks/main.yaml
dk5488: It serves two purposes:
1. **Pre-existing / Non-Kickstart hosts**: If an administrator executes this playbook on an existing KVM host that wasn't provisioned through the kickstart template in this repository, this task ensures the VNC firewall port is opened.
2. **Display changes (Day-2 operations)**: If the user changes `env.z.tigervnc.display` in `all.yaml` later on, kickstart does not re-run, but re-running the playbook will open the newly configured port.

Because `ansible.posix.firewalld` with `state: enabled` is idempotent, it is a no-op if Kickstart already opened the port during installation (similar to how port 80 is handled in both `kvm_host.cfg.j2` and `playbooks/3_setup_kvm_host.yaml`). 

However, if you'd prefer to keep firewall configuration strictly centralized in the kickstart file, let me know and I can remove it.

AmadeusPodvratnik: Correct. In the most cases the KVM host will be created outside of AOP, so leave the code.

AmadeusPodvratnik (approved):
/lgtm

